PR #36, opened 2021-06-09
Allow to opt-in for auto-deployment of an NGINX Ingress Controller + Cert Manager + an Ingress (with TLS support)

rm3l:
**Draft PR**

At the moment, this is an opt-in feature, meaning it does not deploy anything additional unless the environment variables below are set.

It deploys the following resources in the target Kubernetes Cluster:
- NGINX Ingress Controller, using a user-defined Load Balancer IP, if any
- Cert Manager, with a TLS Secret targeting either the production or staging Let's Encrypt ACME server
- an Ingress Resource part of the cosmotech-api Helm Release deployment, in case an FQDN DNS name is provided

For simplicity, this is controlled via few additional environment variables. Might not be the best choice here, but we will ultimately need to consider rewriting all these "simple-but-hard-to-maintain" scripts later on.

Example Usage:
```shell
❯ export NGINX_INGRESS_CONTROLLER_ENABLED=true
export NGINX_INGRESS_CONTROLLER_LOADBALANCER_IP="x.y.z.t"
export CERT_MANAGER_ENABLED=true
export CERT_MANAGER_USE_ACME_PROD=true
export CERT_MANAGER_COSMOTECH_API_DNS_NAME="test.api.example.com"
export CERT_MANAGER_ACME_CONTACT_EMAIL="platform@example.com"
./api/kubernetes/deploy_via_helm.sh \
    latest \
    phoenix  \
    "a-super-secure-password-for-argo-postgresql"  \
    latest  \
    --values /path/to/my/values.yaml
```


vcarluer (approved):
OK for me, nice script :)
One test is red, maybe need to correct it before merge.

PR #49, opened 2021-06-25
runTemplateId in cloud storage path for run template zip resources

vcarluer:
run template id was not in the build storage path for step zip file.

### DIFF
--- a/solution/src/main/kotlin/com/cosmotech/solution/utils/SolutionUtils.kt
+++ b/solution/src/main/kotlin/com/cosmotech/solution/utils/SolutionUtils.kt
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 package com.cosmotech.solution.utils
 
+import com.cosmotech.api.azure.sanitizeForAzureStorage
 import com.cosmotech.solution.domain.RunTemplateHandlerId
 
 fun getCloudPath(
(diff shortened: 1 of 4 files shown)


solution/src/main/kotlin/com/cosmotech/solution/utils/SolutionUtils.kt
```diff
@@ -7,7 +7,8 @@ import com.cosmotech.solution.domain.RunTemplateHandlerId
 fun getCloudPath(
     organizationId: String,
     solutionId: String,
-    handlerId: RunTemplateHandlerId
+    runTemplateId: String,
+    handlerId: RunTemplateHandlerId,
 ): String {
-  return "${organizationId}/${solutionId}/${handlerId}.zip"
+  return "${organizationId}/${solutionId}/${runTemplateId}/${handlerId}.zip"
```
rm3l, line 13: Not sure how this is handled later on, but to be consistent with what is done in the [file upload](https://github.com/Cosmo-Tech/cosmotech-api/blob/runtemplatezip_envvar/solution/src/main/kotlin/com/cosmotech/solution/azure/SolutionServiceImpl.kt#L402-L406), maybe this path here needs to be sanitized as well.

rm3l (approved):

solution/src/main/kotlin/com/cosmotech/solution/utils/SolutionUtils.kt
vcarluer: yes thank you I will add this

rm3l:
LGTM

PR #64, opened 2021-07-16
fix(PROD-7888): Map null and un-handled scenario run statuses to a new 'Unknown' Scenario state

rm3l:
This fixes [PROD-7888](https://spaceport.cosmotech.com/jira/browse/PROD-7888) by introducing a new `Unknown` state for Scenarios, used to handle cases where the status returned by Argo is `null` or not known to us.

There are few cases relevant for this:

- when fetching a scenario right after a successful run request, Argo Server returns a Workflow with a `status` field populated with `null` values, which is returned because the Workflow is about to be scheduled, but is not actually yet. This lasts just few seconds.
- a possible temporary communication error when reaching out to the Scenario Run Service for fetching statuses


vcarluer (approved):
Didn't really analyzed your unit test but code is OK for me.

PR #94, opened 2021-09-17
[PROD-8051] Add info about last runs of Parent and Root to a given child Scenario

rm3l:
- [PROD-8051] Add new 'parentLastRun' and 'rootLastRun' fields to the Scenario API
- [PROD-8051] Add tests for this feature
- [PROD-8051] Make 'findAllScenarios' and 'findScenarioById' methods include info about parent and root last runs


vcarluer (approved):
Nice factorization :D

PR #98, opened 2021-09-20
Split workspace files path for spring security + bug correction on scenarioruns paths

vcarluer:


rm3l (commented):
LGTM. Security is definitely an area we should figure out how to test intensively, to catch those kinds of issues earlier.

rm3l:
BTW, I was about to suggest you separate the (unrelated) changes in `CsmSecurityConfiguration.kt` and the ones in the `doc` and `openapi` directories in 2 separate Git commits, to have a cleaner history.
But I just pushed this commit, which now automatically pushes changes to the generated documentation and PlantUML : https://github.com/Cosmo-Tech/cosmotech-api/commit/b9c827c4609a402e8c979080ddd2da3747fe2f86
So make sure to rebase your branch onto `origin/main`.

rm3l (approved):

PR #102, opened 2021-09-23
Authorized user list on Workspace

vcarluer:
White list of UPN / mail users list on workspace.
Validate ownership or assignment to workspace with white list before handling request of user scopes endpoints.
Ownership validation for all admin scoped endpoints.
=> 403 Forbidden
No validation for user scoped endpoints if no users list set.

Note: Cannot use easily MSGraph for oid / upn search & validation because it needs either:
* A new app registration with API Permission of type Application with 'Users Read All' in the tenant of the customer
* All the Workspace assignment operations for a user needs to be done by an operator of the same tenant

### DIFF
--- a/common/src/main/kotlin/com/cosmotech/api/utils/SecurityUtils.kt
+++ b/common/src/main/kotlin/com/cosmotech/api/utils/SecurityUtils.kt
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 package com.cosmotech.api.utils
 
+import com.azure.spring.aad.AADOAuth2AuthenticatedPrincipal
 import java.lang.IllegalStateException
 import org.springframework.security.core.Authentication
 import org.springframework.security.core.context.SecurityContextHolder
@@ -10,6 +11,15 @@ fun getCurrentAuthentication(): Authentication? = SecurityContextHolder.getConte
 
 fun getCurrentUserName(): String? = getCurrentAuthentication()?.name
 
+fun getCurrentUserUPN(): String? =
+    (getCurrentAuthentication()?.principal as? AADOAuth2AuthenticatedPrincipal)?.attributes
+        ?.getOrDefault("upn", null) as?
+        String?
+
 fun getCurrentAuthenticatedUserName() =
     getCurrentUserName()
         ?: throw IllegalStateException("User Authentication not found in Security Context")
+
+fun getCurrentAuthenticatedUserUPN() =
+    getCurrentUserUPN()
+        ?: throw IllegalStateException("User UPN not found in Authentication Principal")
--- a/workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
+++ b/workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
@@ -14,21 +14,17 @@ import com.cosmotech.api.azure.findByIdOrThrow
 import com.cosmotech.api.azure.sanitizeForAzureStorage
 import com.cosmotech.api.events.OrganizationRegistered
 import com.cosmotech.api.events.OrganizationUnregistered
-import com.cosmotech.api.events.UserAddedToWorkspace
-import com.cosmotech.api.events.UserRemovedFromOrganization
-import com.cosmotech.api.events.UserRemovedFromWorkspace
 import com.cosmotech.api.exceptions.CsmAccessForbiddenException
 import com.cosmotech.api.utils.changed
 import com.cosmotech.api.utils.compareToAndMutateIfNeeded
 import com.cosmotech.api.utils.getCurrentAuthenticatedUserName
+import com.cosmotech.api.utils.getCurrentAuthenticatedUserUPN
 import com.cosmotech.organization.api.OrganizationApiService
 import com.cosmotech.solution.api.SolutionApiService
 import com.cosmotech.user.api.UserApiService
-import com.cosmotech.user.domain.User
 import com.cosmotech.workspace.api.WorkspaceApiService
 import com.cosmotech.workspace.domain.Workspace
 import com.cosmotech.workspace.domain.WorkspaceFile
-import com.cosmotech.workspace.domain.WorkspaceUser
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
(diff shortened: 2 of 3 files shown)


common/src/main/kotlin/com/cosmotech/api/utils/SecurityUtils.kt
```diff
@@ -10,6 +11,15 @@ fun getCurrentAuthentication(): Authentication? = SecurityContextHolder.getConte
 
 fun getCurrentUserName(): String? = getCurrentAuthentication()?.name
 
+fun getCurrentUserUPN(): String? =
+    (getCurrentAuthentication()?.principal as? AADOAuth2AuthenticatedPrincipal)?.attributes
+        ?.getOrDefault("upn", null) as?
+        String?
```
rm3l, line 17: The cast to `AADOAuth2AuthenticatedPrincipal` is too Azure-specific, which is not the goal of this generic file.
I suggest using its `org.springframework.security.oauth2.core.OAuth2AuthenticatedPrincipal` base interface methods instead.
```suggestion
(getCurrentAuthentication()?.principal as? OAuth2AuthenticatedPrincipal)?.getAttribute<String>("upn")
```
Or we could consider moving this to the dedicated `com.cosmotech.api.azure` package.

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
```diff
@@ -43,94 +37,101 @@ import org.springframework.stereotype.Service
...
+    } else {
+      logger.debug(
+          "User {} is authorized on Workspace {} - {}",
+          currentUserUPN,
+          workspace.name,
+          workspace.id)
+    }
+  }
 
   override fun findAllWorkspaces(organizationId: String) =
```
rm3l, line 72: Maybe we should consider making `findAllWorkspaces(String)` return Workspaces with `null` users, owned by or containing the authenticated user ?

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
```diff
@@ -43,94 +37,101 @@ import org.springframework.stereotype.Service
...
-  override fun addOrReplaceUsersInOrganizationWorkspace(
-      organizationId: String,
-      workspaceId: String,
-      workspaceUser: List<WorkspaceUser>
-  ): List<WorkspaceUser> {
-    if (workspaceUser.isEmpty()) {
-      // Nothing to do
-      return workspaceUser
-    }
+    cosmosTemplate.upsert("${organizationId}_workspaces", workspace)
```
rm3l, line 109: Great, but we could optimize this by performing this potentially expensive `upsert` operation only if the updated list of users has really changed.

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
```diff
@@ -43,94 +37,101 @@ import org.springframework.stereotype.Service
...
+  }
 
-    val organization = organizationService.findOrganizationById(organizationId)
+  override fun addUsersInWorkspace(
+      organizationId: kotlin.String,
+      workspaceId: kotlin.String,
+      requestBody: kotlin.collections.List<kotlin.String>
+  ): List<kotlin.String> {
```
rm3l, line 116: No need for specifying `kotlin.` and `kotlin.collections` packages.

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
```diff
@@ -43,94 +37,101 @@ import org.springframework.stereotype.Service
...
-    val workspaceUserMap = workspaceUserWithRightNames.associateBy { it.id!! }
-
-    val currentWorkspaceUsers =
-        workspace.users?.filter { it.id != null }?.associateBy { it.id!! }?.toMutableMap()
-            ?: mutableMapOf()
-
-    newUsersLoaded.forEach { (userId, _) ->
-      // Add or replace
-      currentWorkspaceUsers[userId] = workspaceUserMap[userId]!!
+    if (requestBody.isEmpty()) {
```
rm3l, line 123: For performance reasons, I suggest returning right away if `requestBody` and `workspace.users` are identical (regardless of duplicates and regardless of the order of the elements).
This would prevent performing a useless call to `cosmosTemplate.upsert`.

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
```diff
@@ -191,18 +196,6 @@ internal class WorkspaceServiceImpl(
       hasChanged = true
     }
```
rm3l, line 198: Why are `users` not handled in the `updateWorkspace` method ? I think authorized users should also be able to replace the users list as part of a standard `PATCH` request on the Workspace.

workspace/src/test/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImplTests.kt
```diff
@@ -271,6 +265,8 @@ class WorkspaceServiceImplTests {
     confirmVerified(cosmosTemplate)
   }
 
+  /*
+  Test cannot be perform for now since update checks now for JWT security context
```
rm3l, line 269: Why is this test commented ? It should be possible to adjust it (and eventually test the new behaviors) by mocking module-wide Kotlin functions added in `SecurityUtils.kt`.
See some examples in `ScenarioServiceImplTests`, where `Securityutils::getCurrentAuthentication` is [mocked before each test](https://github.com/Cosmo-Tech/cosmotech-api/blob/main/scenario/src/test/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImplTests.kt#L105-L108) and [unmocked after each test](https://github.com/Cosmo-Tech/cosmotech-api/blob/main/scenario/src/test/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImplTests.kt#L113-L116)

rm3l (changes requested):

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: + [] + role Platform.Admin

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: todo

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: todo

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: todo

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: Refacto needed: Ownership tested at the begining of code: ownerId test no more needed.
=> Remove 'ownerId' and 'users' from mutation excluded fields for direct mutation.
FindSolutionById will throw 404 if not fount which is fine.

workspace/src/test/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImplTests.kt
vcarluer: todo

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: dones

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: done

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: dones

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: dones

workspace/src/main/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImpl.kt
vcarluer: done

workspace/src/test/kotlin/com/cosmotech/workspace/azure/WorkspaceServiceImplTests.kt
vcarluer: done

PR #116, opened 2021-10-12
Versioned storage class prod 8351

vcarluer:
Use csm.platform.api.version:
set in generated config by helm
used in argo builder


vcarluer:
error, branches created on top of another branch...

PR #137, opened 2021-11-10
De-hardcode Workflows Persistent Volume Claims spec building

rm3l:
At this time, running a ScenarioRun in a local cluster requires manual changes
to `WorkflowBuilders.kt`, like changing the storage class and storage request
for PVCs used by the ScenarioRun Workflows.
This makes the development experience unpractical.

The changes here make the PVCs options configurable, with sensible defaults
that are backward-compatible.
This is anyway a preliminary step for PROD-8043, which aims at ultimately
being able to specify a local storage class with `ReadWriteMany` access mode
capabilities.


mahkoCosmo (approved):

vcarluer (approved):
LGTM

PR #138, opened 2021-11-12
download handler. Factorization of run template validation

vcarluer:
endpoint to download a run template with spring boot starter for Azure storage

### DIFF
--- a/solution/src/main/openapi/solutions.yaml
+++ b/solution/src/main/openapi/solutions.yaml
@@ -419,7 +419,7 @@ paths:
                 description: Used only with `application/zip` responses
                 example: inline; filename="handler.zip"
         "404":
-          description: the Workspace file specified is unknown or you don't have access to it
+          description: the Run Template Handler file specified is unknown or you don't have access to it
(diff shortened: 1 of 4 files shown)


solution/src/main/openapi/solutions.yaml
```diff
@@ -372,6 +372,54 @@ paths:
...
+                format: binary
+          headers:
+            Content-Disposition:
+              schema:
+                type: string
+                default: inline; filename="handler.zip"
+                description: Used only with `application/zip` responses
+                example: inline; filename="handler.zip"
+        "404":
+          description: the Workspace file specified is unknown or you don't have access to it
```
rm3l, line 422: Maybe we could change the error description to something more meaningful:
```suggestion
          description: the Run Template Handler file specified is unknown or you don't have access to it
```

rm3l (commented):
Slight change requested regarding the error message in the OpenAPI definition.
Otherwise, LGTM.

rm3l (approved):
LGTM

PR #139, opened 2021-11-12
Search dataset by tag

vcarluer:
mandatory tag list parameter
AND operator

### DIFF
--- a/dataset/src/main/kotlin/com/cosmotech/dataset/azure/DatasetServiceImpl.kt
+++ b/dataset/src/main/kotlin/com/cosmotech/dataset/azure/DatasetServiceImpl.kt
@@ -14,6 +14,7 @@ import com.cosmotech.api.events.ConnectorRemovedForOrganization
 import com.cosmotech.api.events.OrganizationRegistered
 import com.cosmotech.api.events.OrganizationUnregistered
 import com.cosmotech.api.exceptions.CsmAccessForbiddenException
+import com.cosmotech.api.exceptions.CsmClientException
 import com.cosmotech.api.utils.changed
 import com.cosmotech.api.utils.compareToAndMutateIfNeeded
 import com.cosmotech.api.utils.getCurrentAuthenticatedUserName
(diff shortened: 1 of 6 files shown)


dataset/src/main/kotlin/com/cosmotech/dataset/azure/DatasetServiceImpl.kt
```diff
@@ -159,6 +161,32 @@ internal class DatasetServiceImpl(
     TODO("Not yet implemented")
   }
 
+  override fun searchDatasets(organizationId: String, datasetSearch: DatasetSearch): List<Dataset> {
+    var searchList: Set<Dataset> = setOf()
+    if (datasetSearch.datasetTags == null) {
```
rm3l, line 166: Since we expect `datasetTags` to be filled, I suggest we mark it as `required` right in the OpenAPI definition, so a non-nullable Kotlin type is generated.

dataset/src/main/kotlin/com/cosmotech/dataset/azure/DatasetServiceImpl.kt
```diff
@@ -159,6 +161,32 @@ internal class DatasetServiceImpl(
...
+      throw CsmClientException("Dataset tag list is mandatory for search")
+    }
+
+    datasetSearch.datasetTags?.forEach { tag ->
```
rm3l, line 170: As seen together, it would be more efficient, performance-wise, to issue a single SQL Query using logical `AND` operators, instead of `N` queries.

rm3l: Also, I think transforming `datasetTags` into a `Set` would prevent useless operations against the remote data store.

rm3l (changes requested):

dataset/src/main/kotlin/com/cosmotech/dataset/azure/DatasetServiceImpl.kt
rm3l: Also, I think transforming `datasetTags` into a `Set` would prevent useless operations against the remote data store.

dataset/src/main/kotlin/com/cosmotech/dataset/azure/DatasetServiceImpl.kt
vcarluer: done

dataset/src/main/kotlin/com/cosmotech/dataset/azure/DatasetServiceImpl.kt
vcarluer: done

rm3l (approved):

PR #142, opened 2021-11-15
Multiple dataset ids in scenario parameter value

vcarluer:
Handle directly value with format "[DATASETID1,DATASETID2...]" in scenario parameter value
=> No 'isArray' information on solution parameter or dedicated 'arrayValue' in scenario parameter value
=> Direct array in string format and parsing when building workflow

### DIFF
--- a/scenariorun/src/main/kotlin/com/cosmotech/scenariorun/dataset/DatasetUtils.kt
+++ b/scenariorun/src/main/kotlin/com/cosmotech/scenariorun/dataset/DatasetUtils.kt
@@ -8,6 +8,7 @@ import com.cosmotech.connector.domain.Connector
 import com.cosmotech.dataset.api.DatasetApiService
 import com.cosmotech.dataset.domain.Dataset
 import com.cosmotech.scenario.domain.Scenario
+import com.cosmotech.scenario.domain.ScenarioRunTemplateParameterValue
 import com.cosmotech.scenariorun.FETCH_PATH_VAR
 import com.cosmotech.scenariorun.getCommonEnvVars
 import com.cosmotech.scenariorun.resolvePlatformVars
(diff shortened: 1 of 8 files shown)


scenariorun/src/main/kotlin/com/cosmotech/scenariorun/dataset/DatasetUtils.kt
```diff
@@ -42,23 +43,54 @@ internal fun findDatasetsAndConnectors(
...
+          datasets,
+          connectors,
+      )
+    }
+  }
+}
+
+internal fun getDatasetIdListFromValue(paramValue: String): List<String> {
+  if (paramValue.startsWith("[")) {
+    val datasetIds = paramValue.substring(1, paramValue.length - 1)
```
rm3l, line 87: This will throw an Index-out-of-bounds exception if `paramValue` is intentionally set to the one-character `[` String. Perhaps you could also make sure `paramValue` minimum length is 2.
This is actually what you are trying to achieve here, i.e. extract everything between `[` and the last character.

rm3l (changes requested):

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/dataset/DatasetUtils.kt
vcarluer: Done with validation of last char = ']' + unit test

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/dataset/DatasetUtils.kt
```diff
@@ -42,23 +44,57 @@ internal fun findDatasetsAndConnectors(
...
+          connectors,
+      )
+    }
+  }
+}
+
+internal fun getDatasetIdListFromValue(paramValue: String): List<String> {
+  if (paramValue.startsWith("[")) {
+    if (!(paramValue.last().equals(']'))) {
+      throw CsmClientException("Malformed dataset id list: must start with [ and end with ]")
```
rm3l, line 89: I suggest you include the value of `paramValue` in the exception message, so clients can know exactly which value is malformed.

rm3l (changes requested):

PR #154, opened 2021-11-26
[PROD-7420] Check Azure Data Explorer when computing Scenario and ScenarioRun states

rm3l:
## What problem this PR solves
This adds the ability to inspect the data ingestion status in Azure Data Explorer (ADX) any time a Scenario or ScenarioRun state is requested. Actually, such checks are done only if the ScenarioRun Workflow is done in Kubernetes.
Prior to this, Scenario Run states reported by the API were just a mapping of the Argo Workflow execution status. But in practice, once a Scenario Run Workflow is done, probe measures data might not be available right away in Azure Data Explorer (ADX), due to some ingestion latency.

This builds upon the fact that the Simulator Engine is able to send an additional summary message at the end of the simulation to a given Event Hub (controlled by a `CSM_CONTROL_PLANE_TOPIC`  environment variable). By storing such summary info in a dedicated ADX table (aptly named `SimulationTotalFacts`), and checking the expected number of messages against the actual probes measures row count, we can determine whether all data has been ingested or not.

## Edge Cases and mitigation considered
Note that there are few edge cases that have been somehow mitigated.

### Handling data ingestion latency
Depending on several factors (like the ADX cluster sizing or simply network conditions), we might have some latency between EventHub and ADX. This needs to be handled since we might be in a situation where there are no ingestion failures reported and no records neither in the `SimulationTotalFacts` table nor in the `ProbesMeasures` table. This should not be considered an error.

To alleviate this case, we decided to return a transient state for any state request received in the last X seconds after the scenario run workflow end time, where X is an API Configuration Property arbitrarily set by default to 10. The new configuration property is `csm.platform.data-ingestion.waiting-time-before-ingestion-seconds`.

### Handling potential data ingestion failures
We might have data ingestion failures at any time. It is possible to check for such failures (using the ".show ingestion failures" query), but at the time of writing, there is nothing to match a failure record with a given Simulation.

We decided to check for any failures that occurred in the last X minutes after the scenario run workflow end time, where X is an API Configuration property defaulting to 5. The new configuration property is `csm.platform.data-ingestion.ingestion-observation-window-to-be-considered-a-failure-minutes`.

This leads to other issues, unresolved at this time:

- As depicted [here](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/management/ingestionfailures#show-ingestion-failures), ADX Ingestion failures have a retention period of 14 days. So if a failing scenario is not checked once during 14 days, its state will not be computed correctly. **NOTE**: To alleviate this retention period situation, we decided to persist Scenario Run states in Cosmos DB once they reach a terminal state like `Successful` or `Failed`. This way, we won't need to check ADX again past those 14 days. But again, this requires requesting the Scenario Run state at least once during this retention period of 14 days, until it reaches a terminal state. This might be mitigated by checking the Scenario Run state in a background job.

- When running multiple scenarios at the same time, a single data ingestion might aggregate data from several messages. If any of those messages fails to be handled, then the data ingestion would be reported as Failed. This could potentially have some impacts on possibly unrelated scenarios launched at the same moment, depending on when data ingestion is triggered.

### Handling Simulations that have no Consumers
Previously, no summary message was sent at all at the end of the Simulation. This has been fixed in the C++ engine and will be available in Cosmo Tech SDK >= 8.6.0

### Handling the case where the SimulationTotalFacts table is missing from ADX
If for example someone forgot to create this table, there would be no control plane data to rely upon to report data ingestion states.

If the Control Plane Event Hub does not exist, data ingestion state will not be checked. This is in fact the current behavior already in place prior to implementing the work here.

Otherwise, if the Control Plane Event Hub does exist, then an error will be reported to the requester when requesting a Scenario or ScenarioRun state, because this table is mandatory.

### Handling backward compatibility with Simulators built with Cosmo Tech SDK < 8.5.0
In this case, the SimulationTotalFacts table would contain no record for the Simulation Run.

We added 2 new API configuration properties that govern the default behavior to have, if no ingestion failure is detected:
```yaml
csm:
  platform:
    data-ingestion:
      state:
        # state to report if there no ingestion failure detected, and no control plane info,
        # but we have probe measures in ADX.
        # Value should be convertible to DataIngestionState
        # (e.g., InProgress, Successful, Failure, Unknown)
        state-if-no-control-plane-info-but-probe-measures-data: Successful
        # whether to report an error if there is neither control plane info nor probe measures data in ADX
        exception-if-no-control-plane-info-and-no-probe-measures-data: true
```

## Migration strategy
As mentioned above, simulator engines built with Cosmo Tech SDK < 8.5.0 should be handled gracefully.

For a seamless support of this feature, data Ingestion status is checked only if the control plane Event Hub does exist at the time the scenario run request is made. So existing Workspaces could actually create the Control Plane Event Hubs when ready to have data ingestion considered for Scenario Run states.
This allows to safely deploy the changes here, without breaking existing projects.

---
Co-authored-by: @mahkoCosmo


rm3l:
cc @mahkoCosmo

mahkoCosmo (approved):

PR #186, opened 2022-01-05
Delete data from adx prod 8148

mahkoCosmo:


rm3l:
This builds upon #154, which introduced an Azure Data Explorer Client in the code base.
For a better separation of concerns, I will extract and push the Data Explorer client implementation in `main`, as a basis for both #154 and #186.
This way, you would need to rebase your branch onto `main`, and not onto the branch created for #154, which is not merged yet.
This will ease reviewing your changes, and allow to eventually merge your changes independently.

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/DeleteDataByScenarioRunId.kql
```diff
@@ -0,0 +1,3 @@
+.drop extents <|
```
rm3l, line 1: What's the purpose of this file ?

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/azure/ScenarioRunServiceImpl.kt
```diff
@@ -87,6 +87,16 @@ internal class ScenarioRunServiceImpl(
       throw CsmAccessForbiddenException("You are not allowed to delete this Resource")
     }
     cosmosTemplate.deleteEntity("${organizationId}_scenario_data", scenarioRun)
+
+    deleteDataFromScenarioRunId(organizationId, scenarioRun.workspaceKey!!, scenariorunId)
```
rm3l, line 91: Please consider handling the case where data deletion does not succeed in ADX.

rm3l (changes requested):
I just merged #187, which cherry-picks the ADX client introduced in #154.
So please consider rebasing your changes onto `main` instead (resolving any conflicts). This will ease reviewing your PR.

rm3l:
Also please consider adding unit test cases for the new methods.

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/DeleteDataByScenarioRunId.kql
mahkoCosmo: To allow the user to delete the other tables updated with 'policy update'

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/azure/ScenarioRunServiceImpl.kt
mahkoCosmo: C'est deja le cas dans AzureDataExplorerClient.kt non?

rm3l: Nope. I mean, if the call to `deleteDataFromScenarioRunId` errors, there is currently no way to rollback. This would return an error to the client, and leave the system in a state where the ScenarioRun no longer exists but there are still data related to it in ADX. Worse, clients would think the ScenarioRun was not deleted, while the document actually was.

I suggest you implement such rollback mechanism in a simple way, for now simply by catching potential exceptions that might be thrown by `deleteDataFromScenarioRunId` and restoring the ScenarioRun in such exceptional cases.

rm3l: Also, I am curious about the timing of the call to ADX. Is data deletion synchronous or not ? If yes, this might have an impact on the API response time.

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -82,6 +83,29 @@ class AzureDataExplorerClient(private val csmPlatformProperties: CsmPlatformProp
         }
     return healthBuilder.withDetail("baseUri", baseUri).build()
   }
+
+  public fun deleteDataFromScenarioRunId(
```
rm3l, line 87: Kotlin methods and functions are `public` by default.
```suggestion
  fun deleteDataFromScenarioRunId(
```

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -82,6 +83,29 @@ class AzureDataExplorerClient(private val csmPlatformProperties: CsmPlatformProp
...
+      organizationId: String,
+      workspaceKey: String,
+      scenarioRunId: String
+  ) {
+    var databaseName = getDatabaseName(organizationId, workspaceKey)
```
rm3l, line 92: This variable can be declared as `val`. It is not modified anywhere else.
```suggestion
    val databaseName = getDatabaseName(organizationId, workspaceKey)
```

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -82,6 +83,29 @@ class AzureDataExplorerClient(private val csmPlatformProperties: CsmPlatformProp
...
+    try {
+      this.kustoClient.executeToJsonResult(
+          databaseName,
+          """
+        .drop extents <|
+            .show database databaseName extents
```
rm3l, line 98: You meant `${databaseName}`, right? Also I guess you should take care of properly escaping such variable in the Kusto query.

scenariorun/src/test/kotlin/com/cosmotech/scenariorun/azure/ScenarioRunServiceImplTests.kt
```diff
@@ -70,14 +71,16 @@ class ScenarioRunServiceImplTests {
   private lateinit var eventPublisher: CsmEventPublisher
 
   @MockK(relaxed = true) private lateinit var workflowService: WorkflowService
+  @MockK(relaxed = true) private lateinit var azureDataExplorerClient: AzureDataExplorerClient
```
rm3l, line 74: Please consider adding a unit test case, to ensure that your new `deleteDataFromScenarioRunId` method on `AzureDataExplorerClient` really does get called when deleting a ScenarioRun.
This is a method call verification test. See https://mockk.io/#verification-atleast-atmost-or-exactly-times

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
rm3l, line 87: Please consider adding a test case for this method.

rm3l (changes requested):

PR #211, opened 2022-02-07
nginx default tls secret name

vcarluer:


api/kubernetes/deploy_via_helm.sh
```diff
@@ -87,6 +102,8 @@ controller:
     labels:
       networking/traffic-allowed: "yes"
     loadBalancerIP: "${NGINX_INGRESS_CONTROLLER_LOADBALANCER_IP}"
+  extraArgs:
+    default-ssl-certificate: "${NAMESPACE}/${TLS_SECRET_NAME}"
```
rm3l, line 106: Just curious - what happens if the TLS secret does not exist? This may happen for instance if `TLS_CERTIFICATE_TYPE` is set to `none`, which is possible since users may purposely disable TLS.
I guess NGINX Ingress Controller will still work by serving a self-signed certificate.

vcarluer: It has a self signed certificate ant the goal here was to avoid nginx to send it. But you are right, if default-ssl-certificate is incorrect we have to check it serves well the self signed one.

PR #215, opened 2022-02-15
[PROD-8858] Get the kusto request's result in json format then build …

mahkoCosmo:
…a KustoOperationResult from the json string

### DIFF
--- a/common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
+++ b/common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
@@ -9,6 +9,7 @@ import com.cosmotech.api.scenariorun.DataIngestionState
 import com.microsoft.azure.kusto.data.Client
 import com.microsoft.azure.kusto.data.ClientImpl
 import com.microsoft.azure.kusto.data.ClientRequestProperties
+import com.microsoft.azure.kusto.data.KustoOperationResult
 import com.microsoft.azure.kusto.data.auth.ConnectionStringBuilder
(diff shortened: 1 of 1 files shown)


common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -230,19 +231,22 @@ class AzureDataExplorerClient(
...
-                ClientRequestProperties().apply {
-                  timeoutInMilliSec = TimeUnit.SECONDS.toMillis(REQUEST_TIMEOUT_SECONDS)
-                })
-            .primaryResults
+            ClientRequestProperties().apply {
+              timeoutInMilliSec = TimeUnit.SECONDS.toMillis(REQUEST_TIMEOUT_SECONDS)
+            })
+
+    val failureQueryPrimaryResults =
+        KustoOperationResult(failureQueryResultsJsonString, "v1").primaryResults
```
jreynard-code, line 248: Could be interesting to extract the version "v1" here and explain with you use v1 instead of v2.
As far as I know, we do not know if the version "v1" will be supported/used for a long time, so it could be great to use a global constant instead.
Maybe add a link to the implementation for information:
https://github.com/Azure/azure-kusto-java/blob/03aabae864c68bca07519c541a02ffaef4f85c00/data/src/main/java/com/microsoft/azure/kusto/data/KustoOperationResult.java

jreynard-code (changes requested):
Just a little change.
LGTM ;)

jreynard-code (approved):
LGTM

PR #246, opened 2022-04-01
Add new scenario validation status

mahkoCosmo:


common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -285,6 +298,77 @@ class AzureDataExplorerClient(
...
+    ingestionProperties.dataFormat = IngestionProperties.DataFormat.CSV
+    ingestFromStream(streamSourceInfo, ingestionProperties, ingestClient)
+  }
+
+  private fun ingestFromStream(
+      streamSourceInfo: StreamSourceInfo,
+      ingestionProperties: IngestionProperties,
+      ingestClient: IngestClient
+  ) {
+    Thread {
```
jreynard-code, line 350: The Thread block should be reworked:
1 - why do we have 2 try-catch blocks in a raw?
2 - if `result = ingestClient.ingestFromStream(streamSourceInfo, ingestionProperties)` throw an error, result will be `null` which is apparently not permitted in the second try-catch (see `var status = result!!.ingestionStatusCollection[0]`)
...

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -285,6 +298,77 @@ class AzureDataExplorerClient(
...
+          var result: IngestionResult? = null
+          try {
+            result = ingestClient.ingestFromStream(streamSourceInfo, ingestionProperties)
+          } catch (e: java.lang.Exception) {
+            println("Failed to initiate ingestion: " + e.message)
+          }
+          try {
+            var status = result!!.ingestionStatusCollection[0]
+            while (status.status == OperationStatus.Pending) {
```
jreynard-code, line 359: Could you please add another way to break the while loop and handle case when a scenario status is blocked `Pending ?

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -24,6 +36,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.stereotype.Service
 
 private const val REQUEST_TIMEOUT_SECONDS = 30L
+private const val SLEEPING_TIME: Long = 5000
```
jreynard-code, line 39: This parameter should be configurable. Could you please add a property (in csmPlatformProperties) in order to change it without deployment?

scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
```diff
@@ -332,6 +335,13 @@ internal class ScenarioServiceImpl(
       this.findAllScenariosStateOption(organizationId, workspaceId, true)
           .addLastRunsInfo(this, organizationId, workspaceId)
 
+  override fun findAllScenariosByValidationStatus(
+      organizationId: String,
+      workspaceId: String,
+      validationStatus: ScenarioValidationStatus
+  ): List<Scenario> =
+      findAllScenarioByValidationStatus(organizationId, workspaceId, validationStatus.toString())
```
jreynard-code, line 343: Could be interesting to use `validationStatus.value` instead of `validationStatus.toString()`.
toString method can be overridden

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -285,6 +298,77 @@ class AzureDataExplorerClient(
...
+  private fun ingestFromStream(
+      streamSourceInfo: StreamSourceInfo,
+      ingestionProperties: IngestionProperties,
+      ingestClient: IngestClient
+  ) {
+    Thread {
+          var result: IngestionResult? = null
+          try {
+            result = ingestClient.ingestFromStream(streamSourceInfo, ingestionProperties)
+          } catch (e: java.lang.Exception) {
```
jreynard-code, line 354: no need to specify `java.lang.Exception`, `Exception` should be enough

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -285,6 +298,77 @@ class AzureDataExplorerClient(
...
+          }
+          try {
+            var status = result!!.ingestionStatusCollection[0]
+            while (status.status == OperationStatus.Pending) {
+              Thread.sleep(SLEEPING_TIME)
+              status = result.ingestionStatusCollection[0]
+            }
+            println("Ingestion completed")
+            println("Final status: " + status.status)
+          } catch (e: java.lang.Exception) {
```
jreynard-code, line 365: no need to specify `java.lang.Exception`, `Exception` should be enough

scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
```diff
@@ -391,6 +425,15 @@ internal class ScenarioServiceImpl(
     return scenario
   }
 
+  override fun getScenarioValidationStatusById(
+      organizationId: kotlin.String,
```
jreynard-code, line 429: no need to specify `kotlin.String`, `String` should be enough (same below)

scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
```diff
@@ -630,6 +673,12 @@ internal class ScenarioServiceImpl(
       if (datasetListUpdated) {
         publishDatasetListChangedEvent(organizationId, workspaceId, scenarioId, scenario)
       }
+
+      azureDataExplorerClient.ingestScenarioValidationStatus(
+          organizationId,
+          workspace.key,
+          scenarioId,
+          scenario.validationStatus?.toString() ?: ScenarioValidationStatus.Unknown.toString())
```
jreynard-code, line 681: Could be interesting to use validationStatus.value instead of validationStatus.toString(). (and Unknown.value)
toString method can be overridden

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
```diff
@@ -285,6 +298,77 @@ class AzureDataExplorerClient(
...
+
+    ingestionProperties.setAdditionalProperties(mutableMapOf("ignoreFirstRecord" to "true"))
+    val data = "${scenarioId},${scenarioStatus},${ZonedDateTime.now().toLocalDate().toString()}"
+    val inputStream: InputStream = ByteArrayInputStream(StandardCharsets.UTF_8.encode(data).array())
+    val streamSourceInfo = StreamSourceInfo(inputStream)
+    ingestionProperties.dataFormat = IngestionProperties.DataFormat.CSV
+    ingestFromStream(streamSourceInfo, ingestionProperties, ingestClient)
+  }
+
+  private fun ingestFromStream(
```
jreynard-code, line 345: Could you please add simple test in `AzureDataExplorerClientTests` for this method ?

jreynard-code (changes requested):
Great job!
Some changes need to be done.

.detekt/detekt.yaml
```diff
@@ -114,7 +114,7 @@ complexity:
     threshold: 600
   LongMethod:
     active: true
-    threshold: 60
+    threshold: 70
```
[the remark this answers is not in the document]
mahkoCosmo: I put back the original value(60) but there's no more detekt errors!!!

scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
mahkoCosmo: if we've a good reason(complicated string representation such as joining values) to override toString but here we only need a string value

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
mahkoCosmo: Too generic detekt error

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
mahkoCosmo: because the ingestFromStream throw an exception and ingestionStatusCollection throws another one
but we only need the first exception

common/azure/src/main/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClient.kt
mahkoCosmo: don't sure to understand what you meant

common/azure/src/test/kotlin/com/cosmotech/api/azure/adx/AzureDataExplorerClientTests.kt
```diff
@@ -362,4 +363,21 @@ class AzureDataExplorerClientTests {
...
+    every { csmPlatformProperties.azure?.dataWarehouseCluster?.options?.ingestionUri } returns
+        "https://ingest-cluster.location.kusto.windows.net"
+    every { csmPlatformProperties.azure!!.credentials.core.clientId } returns "my-clientId"
+    every { csmPlatformProperties.azure!!.credentials.core.clientSecret } returns "my-clientSecret"
+    every { csmPlatformProperties.azure!!.credentials.core.tenantId } returns "my-tentantId"
+
+    val azureDataExplorerClient =
+        spyk(AzureDataExplorerClient(csmPlatformProperties, eventPublisher))
+
+    assertDoesNotThrow {
```
jreynard-code, line 378: - Could you please add assertion on `ingestClient.ingestFromStream` calls?
- Could you please test that after X retry the loop is closed correctly?
The test right now will never throw an Exception and pass everytime...

jreynard-code (changes requested):
GG, some changes anyway

jreynard-code (changes requested):
GG.
Some changes to do as discussed

jreynard-code (approved):
LGTM ;)

PR #281, opened 2022-05-23
Send scenario and scenariorun meta data to eventhub instead of direct…

mahkoCosmo:
also  in cosmotech-terraform repos adx_table_schema_PROD-7500 branch

### DIFF
--- a/scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
+++ b/scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
@@ -679,7 +679,7 @@ internal class ScenarioServiceImpl(
         publishDatasetListChangedEvent(organizationId, workspaceId, scenarioId, scenario)
       }
 
-      sendScenarioMetaData(organizationId, workspace, scenario)
+      sendScenarioMetaData(organizationId, workspace, existingScenario)
     }
 
     return existingScenario
--- a/scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
+++ b/scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
@@ -287,9 +287,6 @@ internal class ScenarioServiceImpl(
 
     this.handleScenarioDeletion(organizationId, workspaceId, scenario, waitRelationshipPropagation)
     eventPublisher.publishEvent(ScenarioDeleted(this, organizationId, workspaceId, scenarioId))
-
-    val workspace = workspaceService.findWorkspaceById(organizationId, workspaceId)
-    sendScenarioMetaData(organizationId, workspace, scenario)
   }
 
   override fun downloadScenarioData(
@@ -789,6 +786,11 @@ internal class ScenarioServiceImpl(
   ) {
     if (workspace.sendScenarioMetadataToEventHub != true) {
       return
+    }
+
+    if (workspace.useDedicatedEventHubNamespace == false) {
+      logger.warn(
+          "workspace must be configured with useDedicatedEventHubNamespace to true in order to send metadata")
     }
 
     val eventBus = csmPlatformProperties.azure?.eventBus!!
--- a/scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
+++ b/scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
@@ -784,13 +784,14 @@ internal class ScenarioServiceImpl(
       workspace: Workspace,
       scenario: Scenario
   ) {
-    if (workspace.sendScenarioMetadataToEventHub != true) {
+    if (workspace.sendScenarioMetadataToEventHub == false) {
       return
     }
 
     if (workspace.useDedicatedEventHubNamespace == false) {
-      logger.warn(
+      logger.error(
           "workspace must be configured with useDedicatedEventHubNamespace to true in order to send metadata")
+      return
     }
 
     val eventBus = csmPlatformProperties.azure?.eventBus!!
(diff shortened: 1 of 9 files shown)


vcarluer (changes requested):
Tests are failing, can you check that before the review please?

scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
```diff
@@ -778,6 +782,48 @@ internal class ScenarioServiceImpl(
             CosmosItemRequestOptions())
   }
 
+  private fun sendScenarioMetaData(
+      organizationId: String,
+      workspace: Workspace,
+      scenario: Scenario
+  ) {
+    if (workspace.sendScenarioMetadataToEventHub != true) {
```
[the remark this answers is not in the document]
vcarluer: Maybe set the default value in openapi to true

scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
```diff
@@ -778,6 +782,48 @@ internal class ScenarioServiceImpl(
...
+      return
+    }
+
+    val eventBus = csmPlatformProperties.azure?.eventBus!!
+    val eventHubNamespace = "${organizationId}-${workspace.key}".lowercase()
```
vcarluer, line 795: Test here workspace.useDedicatedEventHubNamespace and if false log a WARNING: "workspace must be configured with useDedicatedEventHubNamespace to true in order to send metadata"

workspace/src/main/openapi/workspace.yaml
```diff
@@ -486,6 +486,13 @@ components:
           description: Set this property to true to use a dedicated Azure Event Hub Namespace for this Workspace.
             The Event Hub Namespace must be named \'<organization_id\>-<workspace_id\>\' (in lower case).
             This Namespace must also contain two Event Hubs named \'probesmeasures\' and \'scenariorun\'.
+        sendScenarioMetadataToEventHub:
+          # Deprecated - this option will be removed in the future
```
vcarluer, line 490: Why is it deprecated?

workspace/src/main/openapi/workspace.yaml
```diff
@@ -486,6 +486,13 @@ components:
...
+          type: boolean
+          default: false
```
vcarluer, line 492: Set it to true by default? + useDedicatedEventHubNamespace to true too? => communicate when releasing on this default value change for existing workspace for people to force it to false if needed.

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/azure/ScenarioRunServiceImpl.kt
```diff
@@ -598,4 +609,34 @@ internal class ScenarioRunServiceImpl(
...
     return workflowService.stopWorkflow(findScenarioRunById(organizationId, scenariorunId))
   }
+
+  private fun sendScenarioRunMetaData(
+      organizationId: String,
+      workspace: Workspace,
+      scenarioId: String,
+      simulationRun: String
+  ) {
+    val eventBus = csmPlatformProperties.azure?.eventBus!!
```
vcarluer, line 619: As sendScenarioRunMetadata test:
if (workspace.sendScenarioMetadataToEventHub != true) {

workspace/src/main/openapi/workspace.yaml
mahkoCosmo: sendScenarioMetadataToEventHub default value to true, ok
but don't understand for useDedicatedEventHubNamespace

scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
```diff
@@ -283,6 +287,9 @@ internal class ScenarioServiceImpl(
 
     this.handleScenarioDeletion(organizationId, workspaceId, scenario, waitRelationshipPropagation)
     eventPublisher.publishEvent(ScenarioDeleted(this, organizationId, workspaceId, scenarioId))
+
+    val workspace = workspaceService.findWorkspaceById(organizationId, workspaceId)
+    sendScenarioMetaData(organizationId, workspace, scenario)
```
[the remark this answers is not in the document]
mahkoCosmo: initially it was a scenario validation status feature, so for now I suggest that we do not send metadata for 'delete scenario' and create a new ticket for that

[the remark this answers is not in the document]
vcarluer: Yes it is not needed to publish this when scenario is deleted

scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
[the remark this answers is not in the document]
vcarluer: Yes it will be important to inform correctly everyone about that.

workspace/src/main/openapi/workspace.yaml
vcarluer: ok

workspace/src/main/openapi/workspace.yaml
vcarluer: It is the property just before this one in workspace.yaml.
To get the new event system we need 2 options for the workspace set to true:
useDedicatedEventHubNamespace & sendScenarioMetadataToEventHub.

This 2 options were by default to false which means this new feature is not available by default.
* useDedicatedEventHubNamespace where set to false for backward compatibility but now it is a known feature and we should set it to true to always enable scenario metadata. For now solution must put explicitly useDedicatedEventHubNamespace to true if they want it, after this correction, Solutions that does not want a dedicated event hub will need to set it to false.
* you changed sendScenarioMetadataToEventHub to true by default to it's OK.

There is another way to handle this to think about: handle a platform compatibility version in Workspace. Depending on this version it will change this defaut values.

vcarluer (approved):
Nice

PR #312, opened 2022-07-22
add spring metrics configuration

vcarluer:
The spring configuration part to have metrics exposed for services endpoint with common tags


jreynard-code (approved):
LGTM;)

PR #315, opened 2022-07-27
Upgrade argo workflow version prod 8898

mahkoCosmo:


api/kubernetes/deploy_via_helm-dev.sh
```diff
@@ -29,26 +29,171 @@ if [[ $# -lt 3 ]]; then
...
+primary:
+  podLabels:
+    "networking/traffic-allowed": "yes"
+
+EOF
+
+helm repo add bitnami https://charts.bitnami.com/bitnami
+helm upgrade --install -n ${NAMESPACE} ${POSTGRES_RELEASE_NAME} bitnami/postgresql --version ${POSTGRESQL_VERSION} --values values-postgresql.yaml
+
+export ARGO_POSTGRESQL_SECRET=argo-postgres-config
```
jreynard-code, line 103: Could be interesting to rename `ARGO_POSTGRESQL_SECRET` to `ARGO_POSTGRESQL_SECRET_NAME` or somtehing like this. It would be clearer at read.

api/kubernetes/deploy_via_helm.sh
```diff
@@ -205,19 +218,148 @@ EOF
...
+primary:
+  podLabels:
+    "networking/traffic-allowed": "yes"
+
+EOF
+
+helm repo add bitnami https://charts.bitnami.com/bitnami
+helm upgrade --install -n ${NAMESPACE} ${POSTGRES_RELEASE_NAME} bitnami/postgresql --version ${POSTGRESQL_VERSION} --values values-postgresql.yaml
+
+export ARGO_POSTGRESQL_SECRET=argo-postgres-config
```
jreynard-code, line 269: Could be interesting to rename ARGO_POSTGRESQL_SECRET to ARGO_POSTGRESQL_SECRET_NAME or somehing like this. It would be clearer at read.

api/kubernetes/deploy_via_helm.sh
```diff
@@ -205,19 +218,148 @@ EOF
...
+    accessKeySecret:
+      name: ${MINIO_RELEASE_NAME}
+      key: accesskey
+    secretKeySecret:
+      name: ${MINIO_RELEASE_NAME}
+      key: secretkey
+server:
+  extraArgs:
+  - --auth-mode=server
+  secure: false
```
jreynard-code, line 312: I don't know what the `secure: false` and I'm too afraid to ask :p
Could you please check that this configuration is PROD compliant?

jreynard-code (changes requested):
Some label to change and a question about secure flag.
For the rest, LGTM ;)

api/kubernetes/deploy_via_helm.sh
mahkoCosmo: I think to that to secure the workflows we can configure tls and miinio
http://pipekit.io/blog/production-install-of-argo-workflows
https://blog.argoproj.io/practical-argo-workflows-hardening-dd8429acc1ce

vcarluer (changes requested):
csm-argo/charts can be deleted too I think

api/kubernetes/deploy_via_helm.sh
vcarluer: It is needed to avoid credential between the api and argo

vcarluer (approved):

PR #350, opened 2022-10-11
Excluding tomcat from dependencies

icreated:
It's enough to exclude Tomcat


icreated:
Migration of oktaSpringBootVersion in build.gradle.kts resolved this issue. In the future we need to watch out for this dependency

PR #354, opened 2022-10-13
Fix Rbac bugs: setDefault, removeUser

icreated:
**Error**
On method setDefaultSecurity role comes with double quotes ""
**Cause**
String is converted as json due to mime application/json
**Fix**
replacing string by json {role="editor"} in setSecurityDefault

---

**Error**
Methods setDefaultSecurity & removeAccessControl are not updating DB
**Cause**
using of specific updating methods in these general use cases
**Fix**
removeAccessControl & setDefaultSecurity update with cosmosTemplate directly

---

**Error**
many of methods of ScenarioServiceImpl use a wrong roleDefinition
**Fix**
scenarioPermissions is applied


scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
```diff
@@ -285,7 +286,7 @@ internal class ScenarioServiceImpl(
   override fun deleteAllScenarios(organizationId: String, workspaceId: String) {
     // TODO Only the workspace owner should be able to do this
     val workspace = workspaceService.findWorkspaceById(organizationId, workspaceId)
-    csmRbac.verify(workspace.getRbac(), PERMISSION_WRITE, getCommonRolesDefinition())
+    csmRbac.verify(workspace.getRbac(), PERMISSION_WRITE)
```
vcarluer, line 289: Why we don't pass , scenarioPermissions here?

jreynard-code (approved):
LGTM :)

scenario/src/main/kotlin/com/cosmotech/scenario/azure/ScenarioServiceImpl.kt
vcarluer: it is workspace

PR #400, opened 2022-11-23
Renaming delete twingraph endpoint

icreated:


twingraph/src/main/openapi/twingraph.yaml
```diff
@@ -77,15 +77,15 @@ paths:
             application/json:
               schema:
                 type: string
-  /organizations/{organization_id}/delete/{graphId}:
+  /organizations/{organization_id}/{graph_id}:
```
jreynard-code, line 80: Could you please rename endpoints like this : 
- for delete endpoint: `/organizations/{organization_id}/twingraph/{graph_id}`
- for query endpoint: `/organizations/{organization_id}/twingraph/{graph_id}/query` (and do not graphId in payload)
- for import endpoint: `/organizations/{organization_id}/twingraph/import`

jreynard-code (changes requested):
;)

PR #401, opened 2022-11-24
Restrict Possible queries throught the endpoint

jreynard-code:
Reject queries with edition/deletion key words


twingraph/src/test/kotlin/com/cosmotech/twingraph/utils/TwingraphUtilsTests.kt
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) Cosmo Tech.
+// Licensed under the MIT license.
+package com.cosmotech.twingraph.utils
+
+import io.mockk.mockkStatic
+import io.mockk.unmockkStatic
+import kotlin.test.*
```
mahkoCosmo, line 7: wild import?

mahkoCosmo:
Does only 'ReadOnly' queries have to be restricted?

icreated (approved):
Good test!

PR #407, opened 2022-12-08
Modify TwingraphImport endpoint

jreynard-code:
Change TwinGraphImport structure (endpoint body)
Add Storage-> TwinGraph container
Add mapping between source types and containers

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -39,7 +39,7 @@ group = "com.cosmotech"
 version = scmVersion.version
 
 val kotlinJvmTarget = 17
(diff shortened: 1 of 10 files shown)


build.gradle.kts
```diff
@@ -49,6 +49,7 @@ allprojects {
   apply(plugin = "io.gitlab.arturbosch.detekt")
 
   repositories {
+    mavenLocal()
```
icreated, line 52: to drop it ;)

icreated (changes requested):

vcarluer (approved):
OK

PR #422, opened 2022-12-23
Update the scenario's lastrun if the deleted scenarioRun is the scena…

mahkoCosmo:
…rio's lastrun

### DIFF
--- a/scenariorun/src/main/kotlin/com/cosmotech/scenariorun/azure/ScenarioRunServiceImpl.kt
+++ b/scenariorun/src/main/kotlin/com/cosmotech/scenariorun/azure/ScenarioRunServiceImpl.kt
@@ -34,6 +34,7 @@ import com.cosmotech.api.utils.getCurrentAuthenticatedUserName
 import com.cosmotech.api.utils.toDomain
 import com.cosmotech.scenario.api.ScenarioApiService
 import com.cosmotech.scenario.domain.Scenario
+import com.cosmotech.scenario.domain.ScenarioJobState
 import com.cosmotech.scenariorun.CSM_JOB_ID_LABEL_KEY
 import com.cosmotech.scenariorun.ContainerFactory
 import com.cosmotech.scenariorun.SCENARIO_DATA_DOWNLOAD_ARTIFACT_NAME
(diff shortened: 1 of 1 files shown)


scenariorun/src/main/kotlin/com/cosmotech/scenariorun/azure/ScenarioRunServiceImpl.kt
```diff
@@ -148,6 +151,20 @@ internal class ScenarioRunServiceImpl(
     }
   }
 
+  private fun updateScenarioLastRun(scenarioRun: ScenarioRun) {
+    var scenario =
+        scenarioApiService.findScenarioById(
+            scenarioRun.organizationId!!, scenarioRun.workspaceId!!, scenarioRun.scenarioId!!)
+
+    val lastRunId = scenario.lastRun!!.scenarioRunId
+    if (lastRunId == scenarioRun.id) {
```
vcarluer, line 160: This rule is not clear, it is a guess rule on deletion. Moreover this updateScenarioLastRun function only handle this deletion case so it should be renamed deleteScenarioLastRunEvent- or something like this and remove this test.

vcarluer (changes requested):

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/azure/ScenarioRunServiceImpl.kt
```diff
@@ -149,6 +151,17 @@ internal class ScenarioRunServiceImpl(
     }
   }
 
+  private fun deleteScenarioLastRunEvent(scenarioRun: ScenarioRun) {
+    var scenario =
+        scenarioApiService.findScenarioById(
+            scenarioRun.organizationId!!, scenarioRun.workspaceId!!, scenarioRun.scenarioId!!)
+
+    val lastRunId = scenario.lastRun!!.scenarioRunId
```
jreynard-code, line 159: No need to keep this variable anymore as its not used

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/azure/ScenarioRunServiceImpl.kt
```diff
@@ -149,6 +151,17 @@ internal class ScenarioRunServiceImpl(
...
+    scenario.lastRun = null
```
jreynard-code, line 160: If you remove `lastRun` information, set the scenario status to `Created` also. 
In order to be consistent, if a scenario has no `lastRun`, it means that it hasn't been ran, so the scenario status should be "`Created`" instead of "`Failed`, `Successful`,..."

jreynard-code (changes requested):
Some remarks but LGTM ;)

jreynard-code (approved):
LGTM :)

vcarluer (approved):
OK

PR #463, opened 2023-02-24
Persistent metrics for licensing with down sampling

vcarluer:
default:
7 days data fine grained points
400 days down sampled data points
1 hour down sampling bucket size with max aggregation
No down sample for twingraph (already 1 hour data point)


jreynard-code (approved):
LGTM ;)

PR #478, opened 2023-04-05
Rotation

icreated:
A little method to change metadata: graphName & graphRotation

### DIFF
--- a/twingraph/src/main/kotlin/com/cosmotech/twingraph/api/TwingraphServiceImpl.kt
+++ b/twingraph/src/main/kotlin/com/cosmotech/twingraph/api/TwingraphServiceImpl.kt
@@ -164,6 +164,24 @@ class TwingraphServiceImpl(
     return resultSet.toJsonString()
   }
 
+  override fun updateGraphMetaData(organizationId: String, graphId: String, requestBody: Map<String, String>): Any {
+    val organization = organizationService.findOrganizationById(organizationId)
+    csmRbac.verify(organization.getRbac(), PERMISSION_READ)
+    val metaDataKey = "${graphId}MetaData"
+
+    csmJedisPool.resource.use { jedis ->
+      if (jedis.exists(metaDataKey)) {
+        requestBody
+          .filterKeys { it == "graphName" || it == "graphRotation" }
+          .forEach { (key, value) -> jedis.hset(metaDataKey, key, value)
+        }
+        return jedis.hgetAll(metaDataKey)
+      }
+      throw CsmResourceNotFoundException("No metadata found for graphId $graphId")
+    }
+  }
+
+
   override fun bulkQuery(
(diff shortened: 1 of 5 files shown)


twingraph/src/main/kotlin/com/cosmotech/twingraph/api/TwingraphServiceImpl.kt
```diff
@@ -164,6 +164,35 @@ class TwingraphServiceImpl(
...
 
+  override fun updateGraphMetaData(
+      organizationId: String,
+      graphId: String,
+      requestBody: Map<String, String>
+  ): Any {
+    val organization = organizationService.findOrganizationById(organizationId)
+    csmRbac.verify(organization.getRbac(), PERMISSION_READ)
+
+    if (requestBody.containsKey("graphRotation")) {
```
jreynard-code, line 175: You don't need to check if `requestBody` contains `graphRotation` key as get operator return `null` if not found (see. https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-map/get.html)
You can rewrite code like below:
```
      val graphRotation = requestBody["graphRotation"]?.toInt()
      if (graphRotation == null || graphRotation < 1) {
        throw CsmClientException("GraphRotation should be a positive integer")
      }
```

twingraph/src/main/kotlin/com/cosmotech/twingraph/api/TwingraphServiceImpl.kt
```diff
@@ -164,6 +164,35 @@ class TwingraphServiceImpl(
...
+      val graphRotation = requestBody["graphRotation"]?.toInt()
+      if (graphRotation == null || graphRotation < 1) {
+        throw CsmClientException("GraphRotation should be a positive integer")
+      }
+    }
+
+    val metaDataKey = "${graphId}MetaData"
```
jreynard-code, line 182: Could you make  a String extension instead ? as we certainly use this somewhere else and will certainly reuse it

twingraph/src/main/kotlin/com/cosmotech/twingraph/api/TwingraphServiceImpl.kt
```diff
@@ -164,6 +164,35 @@ class TwingraphServiceImpl(
...
+
+    csmJedisPool.resource.use { jedis ->
+      if (jedis.exists(metaDataKey)) {
+        requestBody.filterKeys { it == "graphName" || it == "graphRotation" }.forEach { (key, value)
```
jreynard-code, line 186: Could you please declare `graphName` and `graphRotation` as global const ?

jreynard-code (changes requested):
LGTM 
Just some remarks ;)

PR #486, opened 2023-04-26
fix null value processing in redis query resultset

lalepee:
null are return as is and send in json response as null json value


icreated (approved):
LGTM

PR #510, opened 2023-06-09
Mig/unexisting solution image not handled properly prod 11181

mahkoCosmo:


scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
```diff
@@ -236,6 +241,7 @@ class ContainerFactory(
         throw IllegalStateException("You cannot start a workspace with no solutionId defined")
     val solution =
         solutionService.findSolutionById(organizationId, workspace.solution.solutionId ?: "")
+    azureContainerRegistryClient.checkSolutionImage(solution.repository!!, solution.version!!)
```
jreynard-code, line 244: Well, you can keep this check on solution image here.
Plus, could you add this check also when updateSolution endpoint is called (before the save obviously) in order to verify that, when solution is updtaing, targeted container exists

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
```diff
@@ -394,6 +403,28 @@ class ContainerFactory(
             ))
   }
 
+  internal fun checkContainerImages(containers: List<ScenarioRunContainer>) {
+    // This method only checks public container registry github
```
jreynard-code, line 407: In my opinion, seems to be useless and time consuming to do this each time a run is launched.
It could be replaced by health endpoint, documentation, process or something else

jreynard-code (changes requested):
Some point to discuss with @bmoreau

solution/src/main/kotlin/com/cosmotech/solution/service/SolutionServiceImpl.kt
```diff
@@ -200,6 +202,10 @@ internal class SolutionServiceImpl(
       hasChanged = true
     }
 
+    if (hasChanged && existingSolution.repository != null) {
```
jreynard-code, line 205: There's no need to check `hasChanged` here.
You only have to check in `solution.repository` and `solution.version` are respectively different from `existingSolution.repository` and `existingSolution.version`

jreynard-code (changes requested):
Some remarks but LGTM

jreynard-code (changes requested):
Some remarks

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/containerregistry/DockerContainerRegistryClient.kt
```diff
@@ -0,0 +1,24 @@
...
+    havingValue = "local",
+    matchIfMissing = true)
+class DockerContainerRegistryClient(private val csmPlatformProperties: CsmPlatformProperties) :
+    RegistryClient {
+
+  override fun getEndpoint() = "Not yet implemented"
+
+  override fun checkSolutionImage(repository: String, tag: String) {
+    // TODO("Not yet implemented")
+  }
```
jreynard-code, line 23: I'm not sure we still need this...
If not, you can remove it.
If it's needed (in a way that we want to check existing solution in a cluster local):
- we need an implementation... maybe have another look on `docker-java` or check if something similar exists
- this class should be in `common`
- an interface should be define and this class should implement it (same thing for `AzureContainerRegistryClient` in `cosmotech-api-azure`)
- interface should be used instead of the azure client

PR #521, opened 2023-06-28
Use scenario's runSizing if provided to run a scenario

mahkoCosmo:
See comments in the ticket in order to test


jreynard-code (approved):
LGTM

PR #538, opened 2023-07-19
Added creationDate to ScenarioRun

neomatamune:
used string date for lisibility


jreynard-code (approved):
LGTM

scenariorun/src/main/openapi/scenariorun.yaml
```diff
@@ -786,6 +786,10 @@ components:
         sdkVersion:
           type: string
           description: the MAJOR.MINOR version used to build the solution solution
+        creationDate:
```
sjoubert, line 789: From the look I took when adding similar fields to the modeling API, it seems other time point fields use a `<action>edAt` name pattern, should this one be consisent too and be `createdAt`?

PR #554, opened 2023-09-05
Add new orchestrator type to Run Templates

neomatamune:
PROD-12221 Support csm-orc in platform API
Definition of Done

    When executing a scenario run from API, user can choose between legacy or Python orchestrator

    Deployed and tested in phoenixdev

Note

    Default behavior to be changed: csm-orc should be the default orchestration mode in API


icreated (approved):
LGTM

PR #573, opened 2023-10-16
fix dataset deletion on scenario deletion

lalepee:
only delete when dataset create by v3 api. Check with the newly created creationDate property which is null for previous version


Leopold-Cramer (approved):
LGTM

Leopold-Cramer (changes requested):
Add tests pls :/

PR #581, opened 2023-11-06
Lcra/add missing security endpoints in solution and dataset services

Leopold-Cramer:


lalepee (changes requested):
Just some minor changes
Verify why check fails.

lalepee (approved):

PR #598, opened 2023-11-24
Set the dependencies correctly if the previous container steps have b…

mahkoCosmo:
…een stacked

### DIFF
--- a/scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
+++ b/scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
@@ -1624,25 +1624,48 @@ internal fun getCommonEnvVars(
   return (minimalEnvVars + commonEnvVars).toMutableMap()
 }
 
+private fun getNewDependenciesIfExist(
+    mergedContainerNames: MutableMap<String, String>,
+    container: ScenarioRunContainer
+): ScenarioRunContainer {
+  var tmp: ScenarioRunContainer = container
+  var dependencies: MutableList<String> = mutableListOf()
+  container.dependencies?.forEach {
+    if (mergedContainerNames.containsKey(it)) {
+      dependencies.add(mergedContainerNames.get(it)!!)
+    }
+  }
+  if (dependencies.isNotEmpty()) {
+    tmp = container.copy(dependencies = dependencies)
+  }
+  return tmp
+}
+
 private fun stackSolutionContainers(
     containers: MutableList<ScenarioRunContainer>
 ): MutableList<ScenarioRunContainer> {
   val stackedContainers: MutableList<ScenarioRunContainer> = mutableListOf()
   var stackedContainer: ScenarioRunContainer? = null
   var stackedIndex = 1
+  val mergedContainerNames: MutableMap<String, String> = mutableMapOf()
   for (container in containers) {
     if (container.solutionContainer != true) {
+      var tmpContainer: ScenarioRunContainer = container
       if (stackedContainer != null) {
         stackedContainers.add(stackedContainer)
+        tmpContainer = container.copy(dependencies = mutableListOf(stackedContainer.name))
         stackedIndex++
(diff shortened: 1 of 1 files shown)


scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
```diff
@@ -1624,25 +1624,48 @@ internal fun getCommonEnvVars(
   return (minimalEnvVars + commonEnvVars).toMutableMap()
 }
 
+private fun getNewDependenciesIfExist(
```
jreynard-code, line 1627: Method name is not explicit, could you please change it ?

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
```diff
@@ -1624,25 +1624,48 @@ internal fun getCommonEnvVars(
...
+  return tmp
+}
+
 private fun stackSolutionContainers(
     containers: MutableList<ScenarioRunContainer>
 ): MutableList<ScenarioRunContainer> {
   val stackedContainers: MutableList<ScenarioRunContainer> = mutableListOf()
   var stackedContainer: ScenarioRunContainer? = null
   var stackedIndex = 1
+  val mergedContainerNames: MutableMap<String, String> = mutableMapOf()
```
jreynard-code, line 1650: Can be rewrite `val mergedContainerNames = mutableMapOf<String, String>()` (same remark for line 1647)

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
```diff
@@ -1624,25 +1624,48 @@ internal fun getCommonEnvVars(
   return (minimalEnvVars + commonEnvVars).toMutableMap()
 }
 
+private fun getNewDependenciesIfExist(
+    mergedContainerNames: MutableMap<String, String>,
+    container: ScenarioRunContainer
+): ScenarioRunContainer {
+  var tmp: ScenarioRunContainer = container
+  var dependencies: MutableList<String> = mutableListOf()
```
jreynard-code, line 1632: Can be rewrite as `val dependencies = mutableListOf<String>()`
 Plus, what append when you have same value several times? 
 How Argo deals with it? Could it be interesting to use `mutableSetOf<String>()` instead ?

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
```diff
@@ -1624,25 +1624,48 @@ internal fun getCommonEnvVars(
...
 private fun stackSolutionContainers(
     containers: MutableList<ScenarioRunContainer>
 ): MutableList<ScenarioRunContainer> {
   val stackedContainers: MutableList<ScenarioRunContainer> = mutableListOf()
   var stackedContainer: ScenarioRunContainer? = null
   var stackedIndex = 1
+  val mergedContainerNames: MutableMap<String, String> = mutableMapOf()
   for (container in containers) {
     if (container.solutionContainer != true) {
+      var tmpContainer: ScenarioRunContainer = container
```
jreynard-code, line 1653: Could you please change `tmpContainer` name for a more explicit name ?

jreynard-code (changes requested):
Some remarks ;)

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
mahkoCosmo: dependencies as list is a legacy code, ref ScenarioRunContainer.kt
In Argo's point of view, it expects a list(not a set) so it doesn't care of duplicated name i suppose

jreynard-code (approved):
LGTM

PR #606, opened 2023-12-13
Add alwaysPull option to solution

mahkoCosmo:


solution/src/main/openapi/solution.yaml
```diff
@@ -749,6 +749,10 @@ components:
         repository:
           type: string
           description: the registry repository containing the image
+        alwaysPull:
+          type: boolean
+          description: the image pull policy
```
sjoubert, line 754: I guess the description is not really accurate. It doesn't actually store the pull policy.

neomatamune (changes requested):
I confirm @sjoubert comments, the description of the new field would gain at being reworded (for example: "should the docker image be always pulled")
And the plantuml file should not have the `User`plus part added to it.

openapi/plantuml/schemas.plantuml
```diff
@@ -649,8 +650,10 @@ Organization -- OrganizationServices : services
 Organization -- OrganizationSecurity : security
 FileUploadValidation -- "0..*" FileUploadMetadata : nodes
 FileUploadValidation -- "0..*" FileUploadMetadata : edges
+User -- "0..*" UserOrganization : organizations
```
sjoubert, line 653: This line should still not be there. Same for line 656 below.

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/service/ScenarioRunServiceImpl.kt
```diff
@@ -497,7 +497,9 @@ class ScenarioRunServiceImpl(
     logger.debug(startInfo.toString())
     val scenarioRunRequest =
         workflowService.launchScenarioRun(
-            startInfo.startContainers, startInfo.runTemplate.executionTimeout)
+            startInfo.startContainers,
+            startInfo.runTemplate.executionTimeout,
+            startInfo.solution.alwaysPull!!)
```
sjoubert, line 502: Is `alwaysPull` nullable with a default value provided in the openapi definition? Do we need `!!`?

sjoubert (commented):
Can we add a minimal test? At least to check if the constructed workflow/template has the correct policy.

scenariorun/src/main/kotlin/com/cosmotech/scenariorun/service/ScenarioRunServiceImpl.kt
mahkoCosmo: without !!, Type mismatch: inferred type is Boolean? but Boolean was expected

sjoubert: Ok, I admit I did not check, I just though that with a default value there was no need for the property to be nullable.

mahkoCosmo:
> Can we add a minimal test? At least to check if the constructed workflow/template has the correct policy.

On a deja 2 tests!!!!

sjoubert:
> > Can we add a minimal test? At least to check if the constructed workflow/template has the correct policy.
> 
> On a deja 2 tests!!!!

I see 2 tests for the `imagePullPolicy` config entry but nothing for the newly introduced `alwaysPull` property of Solution.

sjoubert (approved):

PR #614, opened 2023-12-19
add option datasetCopy to workspace

lalepee:
this option toggle the creation of a copy of dataset when creating a scenario in the workspace


workspace/src/main/openapi/workspace.yaml
Leopold-Cramer, line 1: You could add this new var to openapi examples

Leopold-Cramer (changes requested):

Leopold-Cramer (approved):

PR #624, opened 2024-01-11
While merging solution's containers merge the envVars as well

mahkoCosmo:


scenariorun/src/main/kotlin/com/cosmotech/scenariorun/ContainerFactory.kt
```diff
@@ -1684,7 +1684,14 @@ private fun mergeSolutionContainer(
...
   val modes = "$stackedMode,$containerMode"
-  val envVars = stackedContainer.envVars?.toMutableMap()
+  val envVars =
+      container.envVars
+          ?.let { envVarsOfContainer ->
+            stackedContainer.envVars?.let { envVarsOfstackedContainer ->
+              envVarsOfContainer + envVarsOfstackedContainer
+            }
+          }
+          ?.toMutableMap()
```
jreynard-code, line 1694: Something like this should work :thinking: 
```
  val envVars = container.envVars?.toMutableMap()
  stackedContainer.envVars?.let { envVars?.putAll(it) }
```

jreynard-code (changes requested):
Some remarks

jreynard-code (approved):
LGTM

PR #633, opened 2024-01-26
Fix multiple link/unlink on DatasetService and WorkspaceService

jreynard-code:


dataset/src/main/kotlin/com/cosmotech/dataset/service/DatasetServiceImpl.kt
```diff
@@ -831,9 +815,16 @@ class DatasetServiceImpl(
...
+    val linkedWorkspaceIdList = dataset.linkedWorkspaceIdList
+    if (linkedWorkspaceIdList != null) {
+      if (linkedWorkspaceIdList.contains(workspaceId)) {
+        return dataset
+      } else {
+        dataset.linkedWorkspaceIdList!!.add(workspaceId)
+      }
+    } else {
+      dataset.linkedWorkspaceIdList = mutableListOf(workspaceId)
+    }
```
csm-thu, line 827: Minor suggestion, but readability could probably be improved with something like this:
```
if (dataset.linkedWorkspaceIdList == null) {
  dataset.linkedWorkspaceIdList = mutableListOf(workspaceId)
} else if (dataset.linkedWorkspaceIdList.contains(workspaceId)) {
  return dataset
} else {
  dataset.linkedWorkspaceIdList!!.add(workspaceId)
}
```
(code should be double checked though, I don't know kotlin :stuck_out_tongue: )

Same thing for `addDatasetToLinkedDatasetIdList` in the workspace service

jreynard-code: Done

dataset/src/main/kotlin/com/cosmotech/dataset/service/DatasetServiceImpl.kt
```diff
@@ -863,9 +853,15 @@ class DatasetServiceImpl(
       workspaceId: String
   ): Dataset {
     val dataset = findDatasetById(organizationId, datasetId)
-    dataset.linkedWorkspaceIdList?.remove(workspaceId)
```
lalepee, line 866: doesn't need to check if object exist before remove

or keep it if it's to save a 'save'. but i don't think that's necessary

lalepee (approved):
Some changes proposed

dataset/src/main/kotlin/com/cosmotech/dataset/service/DatasetServiceImpl.kt
jreynard-code: Yup, the `contains` call is here just to not save dataset if it's not necessary.

dataset/src/integrationTest/kotlin/com/cosmotech/dataset/service/DatasetServiceIntegrationTest.kt
```diff
@@ -809,64 +809,82 @@ class DatasetServiceIntegrationTest : CsmRedisTestBase() {
...
   fun `link workspace from dataset`() {
+
+    assertNull(
+        datasetApiService
+            .findDatasetById(organizationSaved.id!!, datasetSaved.id!!)
+            .linkedWorkspaceIdList)
+
+    datasetApiService.linkWorkspace(organizationSaved.id!!, datasetSaved.id!!, workspaceSaved.id!!)
+
+    checkLinkedWorkspaceId()
```
lalepee, line 820: can it have a more specific name.
like checkOnlyOneLinkedWorkspaceId() or better use parameters same name but with a list of id to check if in the list.

workspace/src/integrationTest/kotlin/com/cosmotech/worskpace/service/WorkspaceServiceIntegrationTest.kt
```diff
@@ -400,52 +381,72 @@ class WorkspaceServiceIntegrationTest : CsmRedisTestBase() {
...
+            .linkedDatasetIdList)
+
     workspaceApiService.linkDataset(organizationSaved.id!!, workspaceSaved.id!!, datasetSaved.id!!)
 
-    assertContains(
+    checkLinkedDatasetId()
+
+    workspaceApiService.linkDataset(organizationSaved.id!!, workspaceSaved.id!!, datasetSaved.id!!)
+
+    checkLinkedDatasetId()
```
lalepee, line 396: Same change name

lalepee (approved):
Some enhancing changes proposed

PR #634, opened 2024-02-07
Afos/new run service

neomatamune:


neomatamune:
Followed most review comments, will do a rebase to remove commit 4c1d907 and add more readable commit prior to merge

PR #642, opened 2024-03-06
Lal/fixes

lalepee:


run/src/main/kotlin/com/cosmotech/run/service/RunServiceImpl.kt
```diff
@@ -108,7 +108,10 @@ class RunServiceImpl(
     val run =
         runRepository
             .findBy(organizationId, workspaceId, runnerId, runId)
-            .orElseThrow { throw IllegalArgumentException("Run #$runId not found in #$runnerId") }
+            .orElseThrow {
+              throw IllegalArgumentException(
+                  "Run #$runId not found in #$runnerId. In #$workspaceId, #$organizationId.")
```
neomatamune, line 113: Message feels a bit strange to read.
Would probably need to check for runner existence one way or another instead, and then defer the exception for runner not in workspace to that other check.

neomatamune (changes requested):

neomatamune (approved):
:+1:

PR #643, opened 2024-03-07
Clean deploy test pods automatically on success

sjoubert:
No need to keep successful stopped pods cluttering the cluster


jreynard-code (approved):
LGTM

PR #644, opened 2024-03-12
remove restriction on dataset twincache manipulation

lalepee:
This come for easing the manipulation of dataset twincache in ETL. Upload can now be done for every sourceType not just File type. Trx (transaction) has been removed from crud for twincache entities.


Leopold-Cramer (approved):
C'est vraiment parce que c'est toi

PR #669, opened 2024-03-18
fix generated connector name to simplify redis search

lalepee:
Redis search have a hard time with dot(.)


esasova (approved):
LGTM :+1: (event if I understand nothing :laughing: )

PR #671, opened 2024-03-18
fix generated connector name to simplify redis search

lalepee:
Redis search have a hard time with dot(.)


dataset/src/main/kotlin/com/cosmotech/dataset/service/DatasetServiceImpl.kt
mahkoCosmo, line 1: What's about the existing connector's names?

lalepee: Yes, that's a risk but the name is, in my opinion, complex enough to not conflict with user created connectors name.

vysahjk (approved):

mahkoCosmo (approved):

PR #676, opened 2024-03-29
Use the values returned by redis request to build nodes

mahkoCosmo:


dataset/src/main/kotlin/com/cosmotech/dataset/service/DatasetServiceImpl.kt
```diff
@@ -278,8 +278,9 @@ class DatasetServiceImpl(
       record?.values()?.forEach { element ->
         when (element) {
           is Node -> {
-            val csmGraphEntity = element.toCsmGraphEntity(CsmGraphEntityType.NODE)
```
lalepee, line 281: You rather fix the toCsmGraphEntity function. This function does more than just getProperty.

lalepee (changes requested):
Can you add a test too validate the new behavior ?

mahkoCosmo:
> Can you add a test too validate the new behavior ?

PR #683, opened 2024-04-12
Add RabbitMQ Listener + Queue Management

jreynard-code:


api/src/main/resources/application.yml
```diff
@@ -94,13 +99,24 @@ csm:
...
+        username: "storage_writer_username"
+        password: "storage_writer_password"
       admin:
         username: "storage_admin_username"
         password: "storage_admin_password"
     eventbus:
       host: "localhost"
-      username: "eventbus_admin_username"
-      password: "eventbus_admin_password"
+      port: "5672"
```
sjoubert, line 110: Should `port` be an integer instead of a string?

api/src/main/resources/application.yml
```diff
@@ -94,13 +99,24 @@ csm:
...
+      default-exchange: "csm-exchange"
+      default-queue: "csm"
+      default-routing-key: "csm"
```
sjoubert, line 113: Same remark as in the common repo, this is not something currently configured on the deployment side.

api/src/main/resources/application.yml
```diff
@@ -94,13 +99,24 @@ csm:
       reader:
         username: "storage_reader_username"
         password: "storage_reader_password"
+      writer:
```
neomatamune, line 102: The addition of the writer account could be made in a separate PR instead of having it seems like a dependency to rabbitmq

workspace/src/main/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImpl.kt
```diff
@@ -335,7 +343,10 @@ internal class WorkspaceServiceImpl(
       val pageable: Pageable =
           Pageable.ofSize(csmPlatformProperties.twincache.workspace.defaultPageSize)
       val workspaces = workspaceRepository.findByOrganizationId(organizationId, pageable).toList()
-      workspaceRepository.deleteAll(workspaces)
+      workspaces.forEach {
```
neomatamune, line 346: Why not make use of `WorkspaceService.deleteWorkspace` that cleanly delete the workspace and its data ?

run/src/main/kotlin/com/cosmotech/run/service/amqp/AmqpClientServiceImpl.kt
neomatamune, line 1: Can't seems to find what happen when an `awakeListener` happens while the associated Workspace did not go throught a `WorkspaceCreated` event (during a migration, or after a restart) how are the `Queues` created there ?

neomatamune (commented):
Still have a few questions over the code, can't approve until those are answered

api/src/main/resources/application.yml
jreynard-code: This is part of current Cosmotech RabbitMqConfig, will see next week to:

- remove defaultQueue creation
- change defaultExchange to amq.default (which is here as default when rabbitMQ is deployed)
- remove defaultRoutingKey

workspace/src/main/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImpl.kt
jreynard-code: Because `WorkspaceService.deleteWorkspace` checks RBAC rights.
As we respond to a OrganizationUnredistered event, RBAC should be bypassed.

run/src/main/kotlin/com/cosmotech/run/service/amqp/AmqpClientServiceImpl.kt
jreynard-code: I removed that and we create Queue on runStart only

neomatamune (approved):
Seems ok to me

PR #688, opened 2024-04-22
fix: remove apply on dataset parent connector on sub create

lalepee:
trx function save dataset modification, this case all modification of parent dataset to be persisted


Leopold-Cramer (approved):
alé

PR #694, opened 2024-04-26
add tests for run query endpoint

Leopold-Cramer:


run/src/main/kotlin/com/cosmotech/run/service/RunServiceImpl.kt
```diff
@@ -296,13 +297,14 @@ class RunServiceImpl(
...
       runDataQuery: RunDataQuery
   ): QueryResult {
-
     checkInternalResultDataServiceConfiguration()
-    getRun(organizationId, workspaceId, runnerId, runId)
+    // getRun(organizationId, workspaceId, runnerId, runId)
 
     val runtimeDS =
         DriverManagerDataSource(
-            "jdbc:postgresql://$host:$port/$runId", readerStorageUsername, readerStoragePassword)
```
sjoubert, line 305: Do we go back to hardcode `localhost`, I don't think this is was we want (wrong merge conflict resolve on your part?)

run/src/main/kotlin/com/cosmotech/run/service/RunServiceImpl.kt
```diff
@@ -280,8 +281,8 @@ class RunServiceImpl(
   ): RunData {
 
     checkInternalResultDataServiceConfiguration()
-    val run = getRun(organizationId, workspaceId, runnerId, runId)
```
sjoubert, line 283: If this is really not necessary then just remove the code, don't comment it. But I somehow doubt we want to remove RBAC checks. (Although maybe this is done elsewhere now).

Same a few lines below for another `getRun`

run/src/main/openapi/run.yaml
```diff
@@ -753,7 +753,7 @@ components:
           description: the list of results
           items:
             type: object
-            additionalProperties: true
```
sjoubert, line 756: I'm not sure how an `additionalProperties` to false behave when no specific properties are defined. Is it ok or does it prevent any content (which is probably not what we want)?

sjoubert (changes requested):

jreynard-code (approved):
LGTM

PR #708, opened 2024-05-20
Remove organization services configuration

sjoubert:
These information are not actually used, we only use the properties/values from the config


jreynard-code (approved):
LGTM

PR #715, opened 2024-05-30
Remove state, lastRun, parentLastRun, rootLastRun on Runner + add lastRunId instead

jreynard-code:


runner/src/main/kotlin/com/cosmotech/runner/service/RunnerService.kt
```diff
@@ -77,9 +76,19 @@ class RunnerService(
   }
 
   fun deleteInstance(runnerInstance: RunnerInstance) {
-    if (runnerInstance.isRunning())
-        throw CsmClientException(
-            "Can't delete a running runner : ${runnerInstance.getRunnerDataObjet().id}")
+    val runner = runnerInstance.runner
+
+    if (!runner.lastRunId.isNullOrBlank()) {
```
lalepee, line 81: Can this be moved to a runnerInstance Function instead ?

lalepee (changes requested):

runner/src/main/kotlin/com/cosmotech/runner/service/RunnerService.kt
```diff
@@ -297,5 +279,24 @@ class RunnerService(
       val rbacSecurity = csmRbac.setDefault(this.getRbacSecurity(), role, this.roleDefinition)
       this.setRbacSecurity(rbacSecurity)
     }
+
+    fun delete() {
```
lalepee, line 283: The delete should be in the runnerService class. runnerService Class do action on runnerInstance Class. runnerInstance should only provide status.

runner/src/main/kotlin/com/cosmotech/runner/service/RunnerApiServiceImpl.kt
```diff
@@ -62,8 +61,7 @@ internal class RunnerApiServiceImpl(
   override fun deleteRunner(organizationId: String, workspaceId: String, runnerId: String) {
     val runnerService = getRunnerService().inOrganization(organizationId).inWorkspace(workspaceId)
     val runner = runnerService.getInstance(runnerId).userHasPermission(PERMISSION_DELETE)
```
lalepee, line 63: var runner should be named runnerInstance

lalepee (changes requested):
Sorry some change needed again

neomatamune (approved):

PR #725, opened 2024-06-12
Add missing required s3 helm values

sjoubert:
These are required and need override, but having them here provides documentation


jreynard-code (approved):
LGTM

PR #728, opened 2024-06-12
Return twingraph query results as json list

neomatamune:
Update return type of `dataset/twingraphQuery` to list

This update allows a cleaner explanation of the return type as Cypher queries return list of results

Also this should update the API libraries generated so that there is no need to do user based operation on the return value.

No more need to convert the returned string to working types using external libraries.


jreynard-code (approved):
LGTM

PR #733, opened 2024-06-17
Correctly list all S3 objects using the paging API

sjoubert:
S3Template seems to be limited/bugged and only returns the first page, directly use S3Client to properly list all objects.


jreynard-code (approved):
LGTM

PR #738, opened 2024-06-26
[PROD-13427][PROD-13433] Use workspace id to read the secret

sjoubert:
This was missed in the recent commit that changed the secret name pattern


jreynard-code (approved):
LGTM

PR #741, opened 2024-07-02
Remove argo storage class creation

sjoubert:
We still keep the api config entry to point to a non-default storage class for the worklow.
If one needs a custom storage class it should be created separately, not by the API helm chart.
If we need more control in the helm chart the standard approach seems to take a PVC config as values, not a storage class.


jreynard-code (approved):
LGTM

PR #744, opened 2024-07-09
Improve usage of container registry config

sjoubert:
We need the scheme separate to be able to HTTP query the registry for image existence, but we don't need the scheme for image configuration of the workflow


jreynard-code (approved):
LGTM

PR #745, opened 2024-07-10
[SDCOSMO-1956] Fix Runner child doesn't inherit parent's parameters

jreynard-code:


runner/src/main/kotlin/com/cosmotech/runner/service/RunnerService.kt
```diff
@@ -256,6 +287,57 @@ class RunnerService(
...
+          solution
+              ?.parameterGroups
+              ?.filter { parameterGroup ->
+                runTemplate?.parameterGroups?.contains(parameterGroup.id) == true
+              }
+              ?.flatMap { parameterGroup -> parameterGroup.parameters ?: mutableListOf() }
+      if (!runTemplateParametersIds.isNullOrEmpty()) {
+        val parentParameters = parent.parametersValues?.associate { it.parameterId to it }
+        val runnerParameters = runner.parametersValues?.associate { it.parameterId to it }
+        // TODO Handle default value
```
sjoubert, line 312: Comment to be removed as discussed IRL

runner/src/main/kotlin/com/cosmotech/runner/service/RunnerService.kt
```diff
@@ -256,6 +287,57 @@ class RunnerService(
...
+                    "Parameter $parameterId not found in parent ($parentId) parameters values")
+              }
+            } else {
+              logger.debug(
+                  "Skipping parameter $parameterId, defined neither in the parent nor in this Runner")
+            }
+          } else {
+            logger.debug(
+                "Skipping parameter $parameterId since it is already defined in this Runner")
+          }
```
sjoubert, line 335: I'm not a big fan on nested if-else for error handling. I find sequential if-return easier to read and understand.

runner/src/main/kotlin/com/cosmotech/runner/service/RunnerService.kt
jreynard-code: As last minute conclusion, TODO was modified to contain "Handle default value" context.

sjoubert (approved):

runner/src/main/kotlin/com/cosmotech/runner/service/RunnerService.kt
jreynard-code: Rewrite using filters lambda ;)

PR #747, opened 2024-07-11
[PROD-13480] Remove the 'findConnectorByName' endpoint

sjoubert:
'-' and some other symbols need to be escaped properly to work with a searchable redis field. However, this endpoint was only added to support a previous migration script and is only used once internally. The internal use does not require this to be a proper endpoint, so we can just make it a method. And the internal use will always look for a fixed connector name that does not exhibit the original search issue.


jreynard-code (approved):
LGTM

PR #748, opened 2024-07-12
Dse/fix runstart prod 13530

sellisd:


runner/src/main/openapi/runner.yaml
```diff
@@ -213,9 +213,10 @@ paths:
         "202":
           description: the Run id started
           content:
-            application/json:
```
sjoubert, line 216: This is a working quick fix for the initial issue. We can merge this first, but I'd rather also take the time to design the endpoint properly before the next release. This POST/create endpoint should return the created object not just an id (here a `RunnerInstance`)

jreynard-code (approved):
LGTM

PR #752, opened 2024-07-15
Fix argo volume access mode

sjoubert:
With the new orchestrator we only have a single pod now, so 'Many' access is not required anymore, this should fix azure deployement where we don't force a 'Many' compatible storage anymore


jreynard-code (approved):
LGTM

PR #756, opened 2024-07-17
Rework the getRunLogs endpoint return value

sjoubert:
This simplifies the return data and avoid exposing the internal service responses (loki)


jreynard-code (approved):
LGTM

PR #760, opened 2024-07-17
Fix update solution wrong runtemplateid

jreynard-code:


sjoubert (approved):

sjoubert (approved):
LGTM, pending the determination of whether the api change on the default value is a breaking change (and whether we accept it or not) for the 3.2 target branch.

PR #764, opened 2024-07-18
Fix swagger-ui spec display/config

sjoubert:
Disable code scanning, it conflicts with our own merge spec config


jreynard-code (approved):
LGTM

PR #765, opened 2024-07-19
Improve spring-doc/swagger-ui configuration

sjoubert:
- Disable package scan and solely rely on the merged api file
- Clean individual openapi file spec (remove servers, auth,...)
- Update patch mustache template to the currently used generator version


jreynard-code (approved):
LGTM

PR #769, opened 2024-07-23
Remove createsecret endpoint

jreynard-code:
- Remove createSecretEndpoint
- Add default workspace secret to run container.
  By convention, run container will attach a kubernetes secret (if exists) named organization_id-workspace_id.
  This will allow advanced users to pass information (specially keys, secrets,...) to run container via DevOps actions


doc/.openapi-generator/FILES
```diff
@@ -70,7 +70,6 @@ Models/Workspace.md
 Models/WorkspaceAccessControl.md
 Models/WorkspaceFile.md
 Models/WorkspaceRole.md
-Models/WorkspaceSecret.md
```
sjoubert, line 73: We should actually remove that file from the repo. (and at some point automate this, files are correctly generated when a new one appears but not removed automatically)

sjoubert (changes requested):

sjoubert (approved):

PR #770, opened 2024-07-23
Put back the OAuth config in the OpenAPI spec

sjoubert:
This is required for a proper config of Swagger and the clients


jreynard-code (approved):
LGTM

PR #774, opened 2024-07-24
[PROD-13633] Make startRun return a proper object

sjoubert:
Ideally we should return a RunInstance but this data obejct is defined in the run openapi, not accessible from the runner one. But having a proper object should ease extension if we need to add more data.


jreynard-code (approved):
LGTM

PR #775, opened 2024-07-24
upgrade build test and package workflow

vysahjk:


.github/workflows/build_test_package.yaml
```diff
@@ -0,0 +1,321 @@
...
+          fetch-depth: 0
+      - name: Set up JDK
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: "21"
+
+      - name: retrieve tag name
+        id: refenv
+        uses: Cosmo-Tech/cosmotech-api/.github/actions/retrievetag@ndon/argocd
```
sjoubert, line 245: This should not reference a working branch. Maybe we can extract the retrievetag code as you did, but if we only use it once here, maybe inlining the code is good enough.

.github/containerscan/allowedlist.yaml
```diff
@@ -1,9 +0,0 @@
-general:
```
sjoubert, line 1: Do we have a reason to remove this file? I don't really know what it's for, just checking we didn't remove something used.

sjoubert (changes requested):
* Unless I didn't see something, the doc/lint/openapi/openapi_clients workflow should not have modification in them
* The build_test_package file should not change extension, it makes it difficult to see what changed here

.github/containerscan/allowedlist.yaml
vysahjk: we don't need this configuration anymore, 
I replaced the deprecated action scan-container https://github.com/Azure/container-scan
by
https://github.com/aquasecurity/trivy-action

.github/workflows/build_test_package.yml
```diff
@@ -2,740 +2,320 @@ name: Build, Test and Package
...
-    if: github.event_name != 'pull_request_target'
-
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v4
-
       - name: Set up JDK
-        uses: actions/setup-java@v4
+        uses: actions/setup-java@v3
```
sjoubert, line 26: This version should not be downgraded

.github/workflows/build_test_package.yml
```diff
@@ -2,740 +2,320 @@ name: Build, Test and Package
...
-      - name: Setup Gradle
+          distribution: "temurin"
+          java-version: "21"
+      - name: setup Gradle
         uses: gradle/actions/setup-gradle@v3
-
+        with:
+          gradle-version: 8.8
+          cache-disabled: true
+          cache-read-only: ${{ github.ref != 'refs/heads/main' }}
```
sjoubert, line 35: When I recently updated the gradle use in the actions, I decided to try the default config for the cache and see if we have any issue. And for the gradle version, the action takes care of downloading the correct one based on the repo config so no need to configure that as well.

Since you completely removed the `setup-gradle` task elsewhere I think we can remove it here as well.

.github/workflows/build_test_package.yml
```diff
@@ -2,740 +2,320 @@ name: Build, Test and Package
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: ./gradlew :cosmotech-dataset-api:integrationTest --tests "com.cosmotech.dataset.service.DatasetServiceIntegrationTest" -x spotlessCheck -x openApiValidate
-
-  organization_it:
+        run: |
+          ./gradlew :cosmotech-dataset-api:integrationTest \
+            --tests "com.cosmotech.dataset.service.DatasetServiceIntegrationTest" \
+            -x spotlessCheck \
+            -x openApiValidate
```
sjoubert, line 98: If you want to span the command on multiple line I'd suggest using the `>` multiline YAML marker instead of `|`, this way you also remove the ` \` line breaks.

.github/workflows/build_test_package.yml
```diff
@@ -2,740 +2,320 @@ name: Build, Test and Package
...
+          image-ref: com.cosmotech/cosmotech-api:${{ github.sha }}
+          format: "sarif"
+          severity: "CRITICAL,HIGH"
+          output: "trivy-results.sarif"
 
-      - name: Upload SARIF reports to GitHub Security tab
-        uses: github/codeql-action/upload-sarif@v3
-        if: ${{ always() }}
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v2
```
sjoubert, line 263: Same here, the action version should not be downgraded

.github/workflows/build_test_package.yml
```diff
@@ -2,740 +2,320 @@ name: Build, Test and Package
...
           name: container-image-scan-report
-          path: ${{ steps.scan.outputs.scan-report-path }}
-          retention-days: 3
+          path: "trivy-results.sarif"
+          retention-days: 2
 
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3.3.0
-        if: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
+        uses: docker/login-action@v3.2.0
```
sjoubert, line 276: Same here for the action version downgrade

sjoubert (changes requested):
@jreynard-code @enocquet The helm chart from this repo is no longer uploaded/tagged in the github workflows. I think we need to synchronize (and unify if possible) how the api helm chart is managed for the various use cases, currently we have *3* out-of-sync versions of it (here, the internal helm chart repo and the artifact-hub helm chart repo)

sjoubert (approved):
I think we can merge this, but until we discussed the helm charts organization I'll still consider the chart from this repo the "official" one (even if it's the one from internal-helm-charts being used by argocd)

sjoubert (approved):

jreynard-code (approved):
Can you please have a look at this [link](https://github.com/Cosmo-Tech/cosmotech-api/pull/775#issuecomment-2248295805) please?

All other modifications LGTM btw ;)

jreynard-code:
This PR is outdated. Work around CI/CD will be set by a new team in order to be consistent with actual process.

PR #780, opened 2024-07-30
fix list runner when rbac is off

lalepee:
https://cosmo-tech.atlassian.net/browse/SDCOSMO-1936


sjoubert (changes requested):
See comments and requested changes on #781 about generalizing the test code.

sjoubert (approved):

PR #781, opened 2024-07-30
fix list runner when rbac is off

lalepee:
https://cosmo-tech.atlassian.net/browse/SDCOSMO-1936

### DIFF
--- a/runner/src/integrationTest/kotlin/com/cosmotech/runner/service/RunnerServiceIntegrationTest.kt
+++ b/runner/src/integrationTest/kotlin/com/cosmotech/runner/service/RunnerServiceIntegrationTest.kt
@@ -15,6 +15,8 @@ import com.cosmotech.api.rbac.ROLE_EDITOR
 import com.cosmotech.api.rbac.ROLE_NONE
 import com.cosmotech.api.rbac.ROLE_USER
 import com.cosmotech.api.rbac.ROLE_VIEWER
+import com.cosmotech.api.security.ROLE_ORGANIZATION_USER
+import com.cosmotech.api.security.ROLE_PLATFORM_ADMIN
 import com.cosmotech.api.tests.CsmRedisTestBase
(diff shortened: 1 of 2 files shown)


runner/src/integrationTest/kotlin/com/cosmotech/runner/service/RunnerServiceIntegrationTest.kt
```diff
@@ -199,20 +201,21 @@ class RunnerServiceIntegrationTest : CsmRedisTestBase() {
 
   @Test
   fun `test CRUD operations on Runner as Platform Admin`() {
-    every { getCurrentAuthenticatedRoles(any()) } returns listOf("Platform.Admin")
+    every { getCurrentAccountIdentifier(any()) } returns "random_user_with_patform_admin_role"
+    every { getCurrentAuthenticatedRoles(any()) } returns listOf(ROLE_PLATFORM_ADMIN)
 
     logger.info("should create a second Runner")
-    val runner2 =
+    val runner3 =
```
sjoubert, line 208: There are at least 2 other places in this test that should be updated form 2 to 3 (the log above `a second Runner` and the variable below `runnerSaved2`.
But the best thing to do is probably to update the test to be less tied to what is done exactly in the other tests:
* list and count the current number of runner at the start of the test
* do the test and assert there is `+ 1` runner after the create and back to the original count after the delete
* genericize logs (e.g. "create another Runner",...) and names ("AnotherRunner" or whatever)

sjoubert (changes requested):

runner/src/integrationTest/kotlin/com/cosmotech/runner/service/RunnerServiceIntegrationTest.kt
lalepee: it's log info mistake, data is wipe every test. I'll change log to match what's actually done.

sjoubert (approved):

PR #796, opened 2024-08-23
Add RBAC/Admin management on findAllDatasets

jreynard-code:


dataset/src/integrationTest/kotlin/com/cosmotech/dataset/service/DatasetServiceIntegrationTest.kt
```diff
@@ -142,7 +142,7 @@ class DatasetServiceIntegrationTest : CsmRedisTestBase() {
 
   @BeforeEach
   fun beforeEach() {
-    every { getCurrentAccountIdentifier(any()) } returns CONNECTED_ADMIN_USER
+    every { getCurrentAccountIdentifier(any()) } returns TEST_USER_MAIL
```
lalepee, line 145: You swap user from admin to test, and every thing still work. i love this so much.
One day will have to make this straight.

lalepee (approved):

dataset/src/integrationTest/kotlin/com/cosmotech/dataset/service/DatasetServiceIntegrationTest.kt
jreynard-code: I wrote `makeDataset()` function and adapt other things also to make tests runs correctly after changing this :p

But I agree with you ;)

PR #806, opened 2024-09-25
Update runner parentId on parent deletion

sjoubert:
Update `Runner.parentId` to point to grand parent or null when parent is deleted


jreynard-code (approved):
LGTM

PR #807, opened 2024-09-25
[PROD-13928] Make Runner.datasetList optional again

sjoubert:
Requiring it breaks the updateRunner endpoint, or at minima forces users to provide that field (with current values) even if that's not the field they want to modify.
This should be fixed in a better way when we find the time to rework the API contract and split the data model for creation, update and reading.


jreynard-code (approved):
LGTM

PR #812, opened 2024-10-08
Lal/sdcosmo 1729 acces list scenario

lalepee:


scenario/src/main/kotlin/com/cosmotech/scenario/service/ScenarioServiceImpl.kt
```diff
@@ -548,21 +547,17 @@ internal class ScenarioServiceImpl(
       workspaceId: String,
       scenarioId: String
   ): Scenario {
-    checkInternalResultDataServiceConfiguration()
-    return findScenarioById(
-        organizationId, workspaceId, scenarioId, withLastRun = true, withState = true)
+    return findScenarioById(organizationId, workspaceId, scenarioId, withState = true)
```
jreynard-code, line 550: Why did you remove `checkInternalResultDataServiceConfiguration` method here?

jreynard-code (changes requested):
LGTM
Just  little question ;)

scenario/src/main/kotlin/com/cosmotech/scenario/service/ScenarioServiceImpl.kt
lalepee: And that's why we need review. That's a mistake.

jreynard-code (approved):
LGTM

PR #814, opened 2024-10-17
Upgrade argo-client-java from v3.4.3 to v3.5.11

jreynard-code:
Since Argo Workflows 3.5, listing workflows will also list archived ones.
   See [here](https://argo-workflows.readthedocs.io/en/latest/upgrading/#upgrading-to-v35)
   and [here](https://github.com/argoproj/argo-workflows/issues/10781)
   and [here](https://github.com/argoproj/argo-workflows/blob/release-3.5/ui/src/models/workflows.ts#L542-L560)
   The only way I found to filter archived ones from this list is to filter by following label
   If this label is in the workflow's labels => the workflow is archived
   If not, the workflow is not archived

### DIFF
--- a/run/src/main/kotlin/com/cosmotech/run/workflow/argo/RunArgoWorkflowService.kt
+++ b/run/src/main/kotlin/com/cosmotech/run/workflow/argo/RunArgoWorkflowService.kt
@@ -28,7 +28,6 @@ import io.argoproj.workflow.apis.InfoServiceApi
 import io.argoproj.workflow.apis.WorkflowServiceApi
 import io.argoproj.workflow.models.IoArgoprojWorkflowV1alpha1Workflow
 import io.argoproj.workflow.models.IoArgoprojWorkflowV1alpha1WorkflowCreateRequest
-import io.argoproj.workflow.models.IoArgoprojWorkflowV1alpha1WorkflowList
 import io.argoproj.workflow.models.IoArgoprojWorkflowV1alpha1WorkflowStatus
 import io.argoproj.workflow.models.IoArgoprojWorkflowV1alpha1WorkflowStopRequest
 import java.security.cert.X509Certificate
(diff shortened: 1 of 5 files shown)


run/src/main/kotlin/com/cosmotech/run/workflow/argo/RunArgoWorkflowService.kt
```diff
@@ -232,17 +224,27 @@ internal class RunArgoWorkflowService(
         workflowList =
             newServiceApiInstance<ArchivedWorkflowServiceApi>(this.apiClient)
                 .archivedWorkflowServiceListArchivedWorkflows(
```
sjoubert, line 226: Why are we still looking for archived workflow separately if they are now listed as regular workflows?

run/src/main/kotlin/com/cosmotech/run/workflow/argo/RunArgoWorkflowService.kt
```diff
@@ -256,32 +258,31 @@ internal class RunArgoWorkflowService(
                   null,
                   null,
                   null,
-                  null)!!
+                  null)
+              .items
+              .filter { !it.metadata.labels.containsKey(argoWorkflowArchivedLabel) }
```
sjoubert, line 263: Shouldn't the filter be conditioned by `skipArchive`?

run/src/main/kotlin/com/cosmotech/run/workflow/argo/RunArgoWorkflowService.kt
```diff
@@ -256,32 +258,31 @@ internal class RunArgoWorkflowService(
...
-      throw IllegalStateException(
-          "Scenario run $runId for Organization $organizationId contains a null " +
-              "workflowId or workflowName")
-    }
-    val workflowStatus = getWorkflowStatus(workflowId, workflowName)
+    val workflowName =
+        run.workflowName
+            ?: throw IllegalStateException(
+                "Run $runId for Organization $organizationId contains a null " +
+                    "workflowId or workflowName")
```
sjoubert, line 275: This should only mention `workflowName` now.

sjoubert (approved):

PR #815, opened 2024-10-21
Only pass the PVC storage class if we have one

sjoubert:
This should avoid issues on subsequent chart deployment where the default storage class has been resolved and conflicts with a 'nil' or empty default value


jreynard-code (approved):
LGTM

PR #820, opened 2024-11-07
[PROD-9005] Make sure the parameter's varType are the one from the solution

sjoubert:
Target fix is to rework the api to remove varType from the input data instead of relying on the buggy readOnly config.
In the mean time, consolidate every varType with the reference value from the solution.


jreynard-code (approved):
LGTM

PR #825, opened 2024-11-18
[RMZ-413] upgrade runner ACL lead to server error on ACL dataset management

jreynard-code:


runner/src/integrationTest/kotlin/com/cosmotech/runner/service/RunnerServiceRBACTest.kt
```diff
@@ -2777,15 +2778,15 @@ class RunnerServiceRBACTest : CsmRedisTestBase() {
 
   @TestFactory
   fun `test Dataset RBAC addRunnerAccessControl`() =
```
lalepee, line 2780: Can you add a comment, this behavior is not obvious and it will help understanding.

runner/src/integrationTest/kotlin/com/cosmotech/runner/service/RunnerServiceRBACTest.kt
```diff
@@ -2820,32 +2821,90 @@ class RunnerServiceRBACTest : CsmRedisTestBase() {
...
+                        ?.filter { datasetAccessControl ->
+                          datasetAccessControl.id == TEST_USER_MAIL
+                        }
+                        ?.any { datasetAccessControl -> datasetAccessControl.role == role })
               }
             }
           }
 
+  @Test
+  fun `test Dataset RBAC addRunnerAccessControl with new ACL entry`() {
```
lalepee, line 2856: Could use a comment. this test that a user with a role added on a runner is also add on the dataset. that's a complex behavior.

runner/src/main/kotlin/com/cosmotech/runner/service/RunnerService.kt
```diff
@@ -178,9 +176,16 @@ class RunnerService(
...
-          ?.forEach { newDatasetId ->
-            this.runner.security?.accessControlList?.forEach {
-              this.propagateAccessControlToDataset(newDatasetId, it.id, it.role)
+          ?.mapNotNull {
+            datasetApiService.findByOrganizationIdAndDatasetId(organization!!.id!!, it)
+          }
+          ?.forEach { newDataset ->
+            this.runner.security?.accessControlList?.forEach { roleDefinition ->
+              val newDatasetAcl = newDataset.getRbac().accessControlList
+              if (newDatasetAcl.none { it.id == roleDefinition.id }) {
```
lalepee, line 185: Wow after this i don't what to hear that python List Comprehension are too complex.

lalepee: you should use a function ->

runner/src/main/kotlin/com/cosmotech/runner/service/RunnerService.kt
```diff
@@ -257,26 +275,6 @@ class RunnerService(
               ?: mutableListOf())
     }
 
-    private fun propagateAccessControlToDatasets(userId: String, role: String) {
```
lalepee, line 260: -> this function indeed

lalepee (approved):
Some remarks

lalepee (approved):

PR #827, opened 2024-11-21
fix: check filenames as we do for destination paths

sellisd:


workspace/src/main/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImpl.kt
```diff
@@ -215,6 +215,10 @@ internal class WorkspaceServiceImpl(
       throw IllegalArgumentException("Invalid destination: '$destination'. '..' is not allowed")
     }
     val workspace = getVerifiedWorkspace(organizationId, workspaceId, PERMISSION_WRITE)
+    if (file?.filename?.contains("..") == true || file?.filename?.contains("/") == true) {
```
Leopold-Cramer, line 218: Would it be useful to check for "\" too ?

Leopold-Cramer (approved):

workspace/src/main/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImpl.kt
sellisd: The quote is a possible character for a filename (even though not very useful), and since it cannot be abused to write to a different location I think we should keep it

jreynard-code (approved):
LGTM

PR #830, opened 2024-11-28
fix: Do not trust and send traffic through CORS proxy

sellisd:
To have a nice list of the licenses we were sending users through the codetabs CORS proxy (https://codetabs.com/).


jreynard-code (approved):
LGTM

PR #847, opened 2025-01-15
refactor: update organization API operation IDs to follow REST naming conventions

Leopold-Cramer:
- Changed collection GET endpoints to use 'list' prefix
- Changed single resource GET endpoints to use 'get' prefix
- Changed POST endpoints to use 'create' prefix
- Changed DELETE endpoints to use 'delete' prefix
- Updated security-related operation names for consistency
- Maintained existing operation IDs that already followed conventions


organization/src/main/openapi/organization.yaml
```diff
@@ -14,7 +14,7 @@ tags:
 paths:
   /organizations:
     post:
-      operationId: registerOrganization
+      operationId: createOrganization
       tags:
         - organization
       summary: Register a new organization
```
sjoubert, line 20: The summaries should probably be updated accordingly too. Here it still mentions "registering" an organization. Same for delete/unregister and maybe others.

sjoubert (changes requested):

PR #856, opened 2025-01-24
fix: include all generated client files

sellisd:
When copying the generated client code we did not
copy hidden files so the generated github workflow to run the client tests was never activated.


jreynard-code (approved):
LGTM

PR #859, opened 2025-01-30
fix: repair generated python client code

sellisd:
The openapi-generator-gradle-plugin depends has a transitive dependency on com.samskivert:jmustache:1.15, however the .owasp.dependencycheck.gradle.plugin depends on the more recent 1.16 version so when
both plugins are included the generated python code has syntax errors (see bug issue: https://github.com/OpenAPITools/openapi-generator/issues/20375) To avoid this conflict of plugin dependencies we have to force the resolution of dependencies to
keep the old version. Forcing the dependency also introduces some changes in the typescript client, but fortunately these are new lines that do not seem to have an effect

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -22,6 +22,12 @@ import org.springframework.boot.gradle.tasks.run.BootRun
 // implementations/configurations in a 'buildSrc' included build.
 // See https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources
 
+buildscript {
+    configurations.all {
+        resolutionStrategy.force("com.samskivert:jmustache:1.15")
+    }
+}
+
 plugins {
   val kotlinVersion = "1.9.23"
   kotlin("jvm") version kotlinVersion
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -22,6 +22,12 @@ import org.springframework.boot.gradle.tasks.run.BootRun
 // implementations/configurations in a 'buildSrc' included build.
 // See https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources
 
+buildscript {
```
sjoubert, line 25: Maybe a quick summary of why this explicit dependency requirement is there as well as a link to the openapi underlying issue would be great.

My idea is to clearly indicate that this is a workaround and that in the future is could be ok to remove this and/or to be careful if/when one wants to update this dependency. Having the information in the commit is great but I find it to be very useful to also have the information locally directly in the code.

sjoubert (changes requested):

sjoubert (approved):

jreynard-code (approved):
LGTM

PR #865, opened 2025-02-12
Add constraints on Organization

jreynard-code:
- organization's name is no longer required for updating an organization
- organization's name should have size between 1 and 50 for creation or update


Leopold-Cramer (approved):
LGTM

PR #866, opened 2025-02-14
Initialize MvC tests

jreynard-code:
- Add configuration with TestContainers
- Add simple test on organization creation

### DIFF
--- a/api/src/integrationTest/resources/application-test.yml
+++ b/api/src/integrationTest/resources/application-test.yml
@@ -8,7 +8,7 @@ spring:
       host: "localhost"
       username: "default"
       # Leave it as blank as there's no auth with test container
-      password:
+      password: "test"
       client-type: jedis
 
 csm:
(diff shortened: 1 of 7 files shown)


sjoubert (approved):

api/src/integrationTest/resources/application-test.yml
```diff
@@ -10,11 +10,17 @@ spring:
       # Leave it as blank as there's no auth with test container
       password:
       client-type: jedis
+  main:
+    banner-mode: "off"
+
+logging:
+  level:
+    org.springframework.security: TRACE
```
lalepee, line 18: a bit much, a DEBUG may suffice

run/src/main/kotlin/com/cosmotech/run/metrics/RunMetrics.kt
lalepee, line 1: I don't see how this is used in MVC tests. Is it suppose to be somewhere else ?

lalepee (approved):
Some comments

PR #868, opened 2025-02-18
fix: upgrade to python > 3.8

sellisd:
we can bring up the openapi-generator to v.7.10.0 (but not to latest stable 7.11 without breaking things).

With this minor upgrade to 7.10.0 the generated python client has as minimum requirement python 3.8 instead of 3.7. Both 3.7 and 3.8 versions have reached their end-of-life but at least 3.8 is still available in the github ubuntu latest runners. With this minor change we will be able to avoid a failed github action in the generated python client


sjoubert (approved):
Nitpick: maybe change the commit message to something like "upgrade openapi generator to be compliant with python > 3.8". Before looking at the commit diff I was expecting an upgrade of some python tools/requirements not the openapi generator.

PR #870, opened 2025-02-19
Lcra/refactor solution openapi prod 14210

Leopold-Cramer:


solution/src/main/openapi/solution.yaml
```diff
@@ -24,27 +25,26 @@ paths:
...
             examples:
               Brewery:
                 $ref: '#/components/examples/Brewery'
           application/yaml:
             schema:
-              type: string
-              format: binary
+              $ref: '#/components/schemas/SolutionCreateRequest'
             examples:
               ADTConnector:
```
sjoubert, line 43: I don't think this example should have a different name in YAML vs JSON

solution/src/main/openapi/solution.yaml
```diff
@@ -177,43 +184,66 @@ paths:
...
                 $ref: "#/components/schemas/RunTemplateParameter"
+            examples:
+              Parameters:
+                value:
+                  - id: "param1"
+                    labels:
+                      en: "Parameter 1"
+                      fr: "Paramètre 1"
+                    varType: "string"
+                    defaultValue: "default"
```
sjoubert, line 209: This example and the YAML one just below should probably be factorized and `$ref`ed.
There are others below too: `ParameterGroups`, `RunTemplates` and `RunTemplate` at least.

solution/src/main/openapi/solution.yaml
```diff
@@ -823,64 +1139,66 @@ components:
...
           type: integer
-          description: an optional duration in seconds in which a workflow is allowed to run
+          description: An optional duration in seconds in which a workflow is allowed to run
         deleteHistoricalData:
           $ref: "#/components/schemas/DeleteHistoricalData"
       required:
         - id
-#         - name
+
     DeleteHistoricalData:
```
sjoubert, line 1168: Do we still need this in the API? If I recall correctly this was there to cleanup ADX databases and I'm not sure if we still support this with the new internal RDS. If not this should be removed (here or in a subsequent PR) as well as the endpoint using it cleaned. There also seems to be unused related code left in the implementation (see classes `DeleteHistoricalDataOrganization` and `DeleteHistoricalDataWorkspace`, methods `deleteHistoricalDataWorkspace` and `sendDeleteHistoricalDataWorkspaceEvent`, and maybe more).

solution/src/main/kotlin/com/cosmotech/solution/service/SolutionServiceImpl.kt
```diff
@@ -222,39 +241,39 @@ class SolutionServiceImpl(
...
 
     // solutionId update is allowed but must be done with care. Maybe limit to minor update?
-    var hasChanged =
+
+    val updatedSolution = Solution(
+      id = solutionId,
+      name = solutionUpdateRequest.name ?: existingSolution.name,
+      organizationId = existingSolution.organizationId,
+      ownerId = existingSolution.ownerId,
+      description = solutionUpdateRequest.description,
```
sjoubert, line 257: Similar to previous PR, all fields from the update request are optional, you need to provide the current value as default if the request does not have them. Otherwise we'll overwrite and loose the current value.

runner/src/main/kotlin/com/cosmotech/runner/service/RunnerService.kt
```diff
@@ -298,7 +298,7 @@ class RunnerService(
                 }
         val solution =
             workspace?.solution?.solutionId?.let {
-              solutionApiService.findSolutionById(organization?.id!!, it)
+              solutionApiService.getSolution(organization?.id!!, it)
```
sjoubert, line 301: This is more linked to other PRs but:
- linked to previous one on Organization: `.id!!` should `.id` (here and just below)
- linked to the future one on Runner: the `organization` member can maybe be made non optional? Not sure how a `RunnerService` without organization could work, but I may be wrong. This can probably be checked/done in the dedicated Runner refacto.

sjoubert (changes requested):

solution/src/main/openapi/solution.yaml
```diff
@@ -902,67 +1221,74 @@ components:
...
+          description: An ordered list of Run Template Parameters
           items:
             type: string
-            description: a Run Template Parameter id
+            description: A Run Template Parameter id
       required:
         - id
 #         - labels
-#         - parameters
+#         - parametersy
```
jreynard-code, line 1255: This can be removed

solution/src/main/openapi/solution.yaml
```diff
@@ -902,67 +1221,74 @@ components:
...
-          description: a regex to validate the value
+          description: A regex to validate the value
         options:
           type: object
-          description: freeform options
+          description: Freeform options
           additionalProperties: true
       required:
         - id
 #         - labels
```
jreynard-code, line 1289: This can be removed

jreynard-code (changes requested):
Some remarks :)
Feel free to ask if something needs to be clarified

solution/src/main/openapi/solution.yaml
```diff
@@ -53,19 +53,23 @@ paths:
                 Brewery:
                   $ref: '#/components/examples/Brewery'
         "400":
-          description: Bad request
+          description: Bad request - Invalid request content or parameters
+        "401":
+          description: Authorization failed - incorrect or missing credentials
+        "403":
+          description: Forbidden - insufficient permissions to create solutions
```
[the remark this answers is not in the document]
sjoubert: This could be the occasion to see if we can use [Overlay](https://learn.openapis.org/overlay/) to add the 401, 403 and 500 (maybe 404) responses on all the endpoint automatically.

solution/src/main/openapi/solution.yaml
```diff
@@ -823,64 +1139,66 @@ components:
...
         gitBranchName:
           type: string
-          description: an optional git branch name
+          description: An optional git branch name
         runTemplateSourceDir:
           type: string
-          description: an optional directory where to find the run template source
+          description: An optional directory where to find the run template source
         orchestratorType:
           $ref: "#/components/schemas/RunTemplateOrchestrator"
```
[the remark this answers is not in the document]
sjoubert: Yes, @neomatamune is in the process of finalizing version 2.0 of csm-orc which will remove the legacy orchestrator support.

solution/src/main/openapi/solution.yaml
```diff
@@ -623,192 +740,391 @@ components:
...
-          description: whether or not the run step is done
+          description: Whether or not the run step is done
         postRun:
           type: boolean
-          description: whether or not the post-run step is done
+          description: Whether or not the post-run step is done
         parametersJson:
           type: boolean
-          description: whether or not to store the scenario parameters in json instead of csv
+          description: Whether or not to store the scenario parameters in json instead of csv
```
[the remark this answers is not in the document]
Leopold-Cramer: @neomatamune @lalepee your thoughts on this ?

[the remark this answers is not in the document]
neomatamune: Burn it ! burn it all to the ground !
gif

jreynard-code (changes requested):
Some remark, nothing big to change :sweat_smile: 
GG btw!

jreynard-code (approved):
LGTM ;)

PR #875, opened 2025-02-27
Change dataset ACL management when ACL is modified on Scenario

jreynard-code:
- Main Dataset linked to a scenario are not modified anymore (except when a user is added into linked Scenario ACL)
- Update and remove a user from a Scenario ACL do not change ACL for linked dataset marked as main
- Not marked dataset are modified to match the linked scenario ACL

Note:
Dataset that are marked as main can be shared between several Scenario, update/remove ACL entries on main dataset lead to permission issues


csm-thu:
Removed myself from the reviewers, I don't have the expertise for this, sorry! :smile:

sjoubert (approved):

PR #882, opened 2025-03-10
Remove useless RunTemplate properties

jreynard-code:


run/src/main/kotlin/com/cosmotech/run/RunContainerFactory.kt
```diff
@@ -56,7 +55,6 @@ private const val RUN_TEMPLATE_ID_VAR = "CSM_RUN_TEMPLATE_ID"
 private const val CONTAINER_ORCHESTRATOR_LEGACY_VAR = "CSM_ENTRYPOINT_LEGACY"
```
sjoubert, line 55: Do we still need this one?

sjoubert (approved):

PR #886, opened 2025-03-18
[PROD-13446] Solution.sdkVersion is now retrieved from the docker image

sjoubert:
The field is now read only and is 'computed' at runtime from a label on the docker image.
This requires the solution image heving been built with a SDK version 11.3.0 or newer, but it should be relatively easy to manually add the required label ('com.cosmotech.sdk-version') with older SDK version


jreynard-code (approved):
LGTM

PR #889, opened 2025-03-20
Add CRUD on solution parameter groups

jreynard-code:
### Important:

You can now update solution parameterGroups via solution update endpoint (/organizations/{organization_id}/solutions/{solution_id}.
Now, If no parameter group is specified when a solution update is requested, solution will have empty parameterGroups property after update.

### Additional endpoints:
As we allowed parameterGroups property to be updated when a solution is updated, we added CRUD endpoints for solution parameter group to manage this property easily:

- GET /organizations/{organization_id}/solutions/{solution_id}/parameterGroups
- POST /organizations/{organization_id}/solutions/{solution_id}/parameterGroups
- GET /organizations/{organization_id}/solutions/{solution_id}/parameterGroups/{parameter_group_id}
- PATCH /organizations/{organization_id}/solutions/{solution_id}/parameterGroups/{parameter_group_id}
- DELETE /organizations/{organization_id}/solutions/{solution_id}/parameterGroups/{parameter_group_id}


solution/src/main/openapi/solution.yaml
```diff
@@ -308,54 +308,124 @@ paths:
...
-      operationId: updateSolutionParameterGroups
+    post:
+      operationId: createSolutionParameterGroup
       tags:
         - solution
-      summary: Update solution parameter groups
+      summary: Create a solution parameter group
       requestBody:
-        description: Parameter groups to update
+        description: Parameter groups to create
```
sjoubert, line 317: This should be singular "group", same in the response description

solution/src/main/openapi/solution.yaml
```diff
@@ -308,54 +308,124 @@ paths:
...
+      responses:
+        "200":
+          description: Parameter groups successfully listed
           content:
             application/json:
               schema:
                 type: array
                 items:
                   $ref: '#/components/schemas/RunTemplateParameterGroup'
         "400":
```
sjoubert, line 351: The 400 response does not make sense on a list endpoint that has no request body

solution/src/main/openapi/solution.yaml
```diff
@@ -308,54 +308,124 @@ paths:
...
+        - solution
+      summary: Get details of a solution parameter group
+      responses:
+        "200":
+          description: Parameter groups successfully retrieved
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RunTemplateParameterGroup'
+        "400":
```
sjoubert, line 415: Same here the 400 response should not be there on a request-less endpoint

solution/src/main/openapi/solution.yaml
```diff
@@ -1104,10 +1180,47 @@ components:
...
+        parentId:
+          type: string
+          description: The Run Template Group parent Id
+        parameters:
+          type: array
+          description: An ordered list of Run Template Parameters
+          items:
+            type: string
+            description: A Run Template Parameter id
+      required:
```
sjoubert, line 1204: Should we have more required fields here (or default values to make them non-optional)?
Like `description`, `isTable`, `parameters`, more?

solution/src/main/openapi/solution.yaml
```diff
@@ -1119,12 +1232,41 @@ components:
...
+      description: A Parameter Group Update Request for a Run Template
+      properties:
+        description:
+          type: string
+          description: A description of the parameter group
+        labels:
+          $ref: '#/components/schemas/TranslatedLabels'
+        isTable:
+          type: boolean
+          default: false
```
sjoubert, line 1253: There should not be default values on UpdateRequest fields, otherwise this means that not providing it explicitly will update them to their default value instead of keeping their value.
Same for `parameters`.

solution/src/main/kotlin/com/cosmotech/solution/service/SolutionServiceImpl.kt
```diff
@@ -549,6 +603,42 @@ class SolutionServiceImpl(
...
+        parameterGroups?.groupBy { it.id.lowercase() }?.filterValues { it.size > 1 }?.isNotEmpty()
+            ?: false
+
+    val hasDuplicateRunTemplateId =
+        runTemplates?.groupBy { it.id.lowercase() }?.filterValues { it.size > 1 }?.isNotEmpty()
+            ?: false
+
+    if (hasDuplicateParameterId || hasDuplicateParameterGroupId || hasDuplicateRunTemplateId) {
+      throw IllegalArgumentException(
+          "Several solution parameters or parameter groups or run templates have same id!")
```
sjoubert, line 639: Should we be more specific and have a dedicated exception for each check? This would be more user-friendly than letting the users check all 3 lists for a duplicate.

solution/src/main/openapi/solution.yaml
jreynard-code: I set a default value for :

- isTable (to false)
- parameters (to [] )

sjoubert (approved):

PR #891, opened 2025-03-20
Clean the generated openapi Markdown docs before (re)generation

sjoubert:
The generator won't remove docs for elements that are removed, so we need to clean it manually in order to avoid accumulating old and now irrelevant files


jreynard-code (approved):
LGTM

PR #894, opened 2025-03-21
Support JSON and YAML as input and output of the API

sjoubert:
Make sure we have both support everywhere and that they take the same schemas


jreynard-code (approved):
LGTM

PR #895, opened 2025-03-21
Fix: Task ':spotlessJava' uses this output of task ':cosmotech-api:openApiMarkdownGenerate' without declaring an explicit or implicit dependency

jreynard-code:


sjoubert (approved):
Approved to fix the specific build command issue you have (and that I reproduce). But I'm not very happy with introducing dependencies that I/we don't understand and that, at first glance, should not be there.

jreynard-code:
Problem disappears :man_shrugging: keep this PR for posterity

PR #905, opened 2025-03-24
feat: make run logs recuperation call argo

Leopold-Cramer:
with the run logs removed from loki, we now have to go through argo to get the active run logs in kubernetes and the finished run logs in seaweed


run/src/main/kotlin/com/cosmotech/run/workflow/argo/RunArgoWorkflowService.kt
```diff
@@ -226,25 +224,55 @@ internal class RunArgoWorkflowService(
...
   }
 
-  override fun getRunLogs(run: Run): RunLogs {
+  override fun getRunningLogs(run: Run): String {
     val workflowName = run.workflowName
-    var containersLogs = listOf<RunLogsEntry>()
+    var result = ""
+    val clientBuilder = RestClient.builder()
+    val client = clientBuilder.baseUrl(csmPlatformProperties.argo.baseUri).build()
+    if (workflowName != null) {
```
jreynard-code, line 232: Argo can throw an APIException (HTTP Error code 500).
Could you please add a handler method in `RunExceptionHandling.kt` to format that error properly.

Something like:


```
  @ExceptionHandler(ApiException::class)
  fun handleApiException(exception: ApiException): ProblemDetail {
    var response = ProblemDetail.forStatus(HttpStatus.INTERNAL_SERVER_ERROR)
    val internalServerError = HttpStatus.INTERNAL_SERVER_ERROR
    response.type = URI.create(httpStatusCodeTypePrefix + internalServerError.value())
    response.title = "Argo Workflows has encountered a problem. Please check Argo logs for more details."
    response.detail = exception.message

    return response
  }
```

Doing that, API response will look like that:

```
{
  "type": "https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500",
  "title": "Argo has encountered a problem. Please check Argo logs for more details.",
  "status": 500,
  "detail": "Internal Server Error",
  "instance": "/organizations/o-XXXX/workspaces/w-XXXX/runners/r-XXXX/runs/run-XXXX/logs"
}
```

run/src/main/kotlin/com/cosmotech/run/workflow/argo/RunArgoWorkflowService.kt
```diff
@@ -226,25 +224,55 @@ internal class RunArgoWorkflowService(
     return buildRunStatusFromWorkflowStatus(run, workflowStatus)
   }
 
-  override fun getRunLogs(run: Run): RunLogs {
+  override fun getRunningLogs(run: Run): String {
     val workflowName = run.workflowName
-    var containersLogs = listOf<RunLogsEntry>()
+    var result = ""
+    val clientBuilder = RestClient.builder()
```
jreynard-code, line 230: You can define the RestClient as a class property instead.

Something like this:

`private val argoClient = RestClient.builder().baseUrl(csmPlatformProperties.argo.baseUri).build()`

run/src/main/kotlin/com/cosmotech/run/workflow/argo/RunArgoWorkflowService.kt
```diff
@@ -226,25 +224,55 @@ internal class RunArgoWorkflowService(
...
-            containersLogs =
-                lokiService
-                    .getPodLogs(
-                        csmPlatformProperties.argo.workflows.namespace,
-                        it.id,
-                        it.startedAt ?: Instant.now())
-                    .map { RunLogsEntry(line = it) }
+            val artifact = newServiceApiInstance<ArtifactServiceApi>(this.apiClient)
+            val response =
+                artifact.artifactServiceGetOutputArtifactByUID(run.workflowId, it.id, "main-logs")
```
jreynard-code, line 271: This can be reduced to a 1liner

jreynard-code (changes requested):
Some remarks.
GJ ;)

lalepee (approved):

jreynard-code (approved):
LGTM

PR #918, opened 2025-04-04
Sustaining based on SonarQube warnings

jreynard-code:
- remove harcoded versions in build.gradle.kts files
- merge into allProjects dependencies, redundant dependencies of all sub-projects
- replace if/checks by require/check/requireNotNull
- code Cleaning (naming,...)

### DIFF
--- a/run/src/main/kotlin/com/cosmotech/run/service/RunServiceImpl.kt
+++ b/run/src/main/kotlin/com/cosmotech/run/service/RunServiceImpl.kt
@@ -70,6 +70,7 @@ internal const val TEXT_POSTGRESQL_TYPE = "TEXT"
 internal const val JSON_KEY_WEIGHT = 3
 internal const val JSON_POSTGRESQL_TYPE = "JSONB"
 internal const val CONFLICT_KEY_WEIGHT = JSON_KEY_WEIGHT
+internal const val POSTGRESQL_DRIVER_CLASS_NAME = "org.postgresql.Driver"
 
 internal val jsonTypeMapWeight =
     mapOf(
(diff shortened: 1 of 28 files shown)


run/src/main/kotlin/com/cosmotech/run/service/RunServiceImpl.kt
```diff
@@ -324,8 +326,7 @@ class RunServiceImpl(
     val run = getRun(organizationId, workspaceId, runnerId, runId)
     run.hasPermission(PERMISSION_WRITE)
 
-    if (sendRunDataRequest.data!!.isEmpty())
-        throw IllegalArgumentException("Data field cannot be empty")
+    require(sendRunDataRequest.data!!.isNotEmpty()) { "Data field cannot be empty" }
```
sjoubert, line 329: Should `sendRunDataRequest.data` have been made required in the recent openapi refacto? (Same for `id` just below). 
* If so, maybe this is better done in another PR since this is not the focus of these changes;
* If not, do we want to handle the data-null using `!isNullOrEmpty()` in order to use the error message also for the null case.

workspace/src/main/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImpl.kt
```diff
@@ -252,8 +248,8 @@ internal class WorkspaceServiceImpl(
       overwrite: Boolean,
       destination: String?
   ): WorkspaceFile {
-    if (destination?.contains("..") == true) {
-      throw IllegalArgumentException("Invalid destination: '$destination'. '..' is not allowed")
+    require(destination?.contains("..") != true) {
```
sjoubert, line 251: Is the inverted logic still correct when `destination` is `null`. It was not an issue before but now it will trigger the error, unless `?.` does not behave like I think.

workspace/src/main/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImpl.kt
```diff
@@ -587,14 +582,14 @@ internal class WorkspaceServiceImpl(
         .not()) {
       val username = getCurrentAccountIdentifier(csmPlatformProperties)
       val retrievedAC = workspace.security.accessControlList.firstOrNull { it.id == username }
-      if (retrievedAC != null) {
-        return workspace.copy(
+      return if (retrievedAC != null) {
```
sjoubert, line 585: Same remark for the ternary as in `SolutionServiceImpl.kt`

sjoubert (changes requested):

sjoubert (approved):

PR #289, opened 2022-06-16
Jrey/use common azure libs

jreynard-code:
- Remove common module and submodules
- Extract common module and submodules into separate libraries (https://github.com/Cosmo-Tech/cosmotech-api-azure and https://github.com/Cosmo-Tech/cosmotech-api-common)
- Upgrade kotlinVersion to 1.7

PR #923, opened 2025-04-10
Lcra/change workspace files storage to seaweed prod 14290

jreynard-code:


api/src/integrationTest/resources/application-test.yml
```diff
@@ -105,5 +119,10 @@ csm:
         default-page-size: 5
     rbac:
       enabled: true
+    s3:
+      endpointUrl: "http://localhost:9000"
+      bucketName: "test-bucket"
+      accessKeyId: "s3_username"
+      secretAccessKey: "s3_password"
```
sjoubert, line 126: This is not consistent with the `spring.cloud.aws.s3` content above. Should this be simply removed since the spring config should be sufficient to configure the S3 client?

Same for all the other test application files.

api/src/main/resources/application.yml
```diff
@@ -34,6 +34,19 @@ spring:
       enabled: false
     kubernetes:
       enabled: false
+    aws:
+      credentials:
+        access-key: ${csm.platform.s3.accessKeyId}
+        secret-key: ${csm.platform.s3.secretAccessKey}
+      s3:
+        # We don't need/have a region for our local S3 service but the AWS SDK requires one to be configured
+        region: ${csm.platform.s3.region}
```
sjoubert, line 43: The original/old implementation had a hardcoded `dummy` value here. I don't think we really need to be able to configure that part. But if you want to, should we have the corresponding `region` entry in the API properties. Assuming it is not an error to refer to non existing entries, then this will be empty for now which if I recall doesn't work.

config/application-dev.sample.yml
```diff
@@ -40,6 +40,15 @@ spring:
...
             certificate: "classpath:[fill-this-value].pem" # certificate file
+  cloud:
+    aws:
+      credentials:
+        access-key: "s3_username"
+        secret-key: "s3_password"
+      s3:
+        region: "dummy"
+        path-style-access-enabled: true
+        endpoint: "http://localhost:9000"
```
sjoubert, line 51: These have default values in `application.yaml`, should they be put in the dev sample file? The application `s3` entry should be the only block useful by default.

workspace/src/test/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImplTests.kt
```diff
@@ -159,8 +152,6 @@ class WorkspaceServiceImplTests {
         workspaceServiceImpl.createWorkspaceFile(ORGANIZATION_ID, WORKSPACE_ID, file, false, null)
     assertNotNull(workspaceFile.fileName)
     assertEquals("my_file.txt", workspaceFile.fileName)
-    assertTrue(
```
sjoubert, line 162: All the "assertTrue/False" on the blob existence have been removed. Do we want / can we check if the S3 blob is actually there instead? Not sure if this is easy or a good idea though.

sjoubert (changes requested):

api/src/integrationTest/resources/application-test.yml
jreynard-code: In fact, we cannot removed these properties because in CsmPlatformProperties `s3` is **not** marked as nullable.
Removing this from `application-*-test.yaml` will lead to an error when Spring will try to build the applicationContext. In particular, when Spring will try to build CsmPlatformProperties bean because s3 is expected in bean's definition.

api/src/main/resources/application.yml
jreynard-code: I did this in a "last-time minute" change in this PR [#289](https://github.com/Cosmo-Tech/cosmotech-api-common/pull/289)

config/application-dev.sample.yml
jreynard-code: I agree ;)

workspace/src/test/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImplTests.kt
jreynard-code: I removed these checks because previously workspace files CRUD were done "directly on filesystem" and checks were possible in unit tests.
Right now, we use an external service, these checks should be covered by integrationTests (which is the case)

sjoubert (approved):

PR #944, opened 2025-04-17
Dse/aggregate reports prod 14264

sellisd:
A quick approach to get an overview of the build reports in the form of html while waiting for a better solution that links the generated reports with a dedicated server (e.g. sonarqube)

While waiting for something better here is a quick win: I added a new gradle task that creates an html page with links towards soe of the generated reports. I included only reports that are copied from the subprojects in the links to avoid adding multiple tasks. If this direction is helpful we can incorporate the remaining generated reports (e.g. test results) and make the generation part of the CI/CD publishing the page in gh-pages.


jreynard-code (changes requested):
LGTM

jreynard-code (approved):
LGTM

PR #954, opened 2025-05-13
docs: update scanner config docs

sellisd:
Updated the comments in the code scanner properties file to bring it up to date. It now uses a template so it should be the same across all projects that rely on the same scanner


jreynard-code (approved):
LGTM

PR #1017, opened 2025-07-01
Ensure unique file names for dataset parts during creation and update operations

jreynard-code:
- Added validation to enforce unique file names in `DatasetServiceImpl`.
- Introduced integration tests for dataset creation and update with duplicate file names.

### DIFF
--- a/dataset/src/integrationTest/kotlin/com/cosmotech/dataset/service/DatasetServiceIntegrationTest.kt
+++ b/dataset/src/integrationTest/kotlin/com/cosmotech/dataset/service/DatasetServiceIntegrationTest.kt
@@ -409,6 +409,67 @@ class DatasetServiceIntegrationTest() : CsmTestBase() {
   }
 
   @Test
+  fun `test createDataset with two dataset part with same name`() {
+
+    val customerPartName = "Customers list"
+    val customerPartDescription = "List of customers"
+    val customerPartTags = mutableListOf("part", "public", "customers")
+    val customerPartCreateRequest =
+        DatasetPartCreateRequest(
+            name = customerPartName,
+            sourceName = CUSTOMER_SOURCE_FILE_NAME,
+            description = customerPartDescription,
+            tags = customerPartTags,
+            type = DatasetPartTypeEnum.File)
+
+    val datasetName = "Shop Dataset"
+    val datasetDescription = "Dataset for shop"
+    val datasetTags = mutableListOf("dataset", "public", "shop")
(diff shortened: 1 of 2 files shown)


dataset/src/integrationTest/kotlin/com/cosmotech/dataset/service/DatasetServiceIntegrationTest.kt
```diff
@@ -408,6 +408,67 @@ class DatasetServiceIntegrationTest() : CsmTestBase() {
     assertEquals(DatasetPartTypeEnum.File, inventoryCreatedDatasetPart.type)
   }
 
+  @Test
+  fun `test createDataset with two dataset part with same name`() {
```
csm-thu, line 412: The test name could be more explicit: "test createDataset with two dataset parts with the same name must fail" :wink:

csm-thu (approved):

dataset/src/main/kotlin/com/cosmotech/dataset/service/DatasetServiceImpl.kt
```diff
@@ -682,6 +682,12 @@ class DatasetServiceImpl(
       "Number of files must be equal to the number of parts if specified. " +
           "${files.size} != ${datasetCreateRequest.parts?.size}"
     }
+    require(
+        files.groupingBy { it.originalFilename }.eachCount().filter { it.value > 1 }.isEmpty()) {
+          "Part File names should be unique during dataset creation. " +
```
sjoubert, line 687: It's not obvious to me which type of "part" file this refers to, even with the code so I guess this may be confusing to users.
If I'm not mistaken you are checking the unicity of the `Multipart file names ...` not the `Dataset part names ...`. In addition to the start of the error message, I think the 2 data fields could be made more explicit. Something like `Multipart file names: ...` and `Dataset parts source names: ...`

Same for the update query below.

sjoubert (changes requested):

dataset/src/integrationTest/kotlin/com/cosmotech/dataset/service/DatasetServiceIntegrationTest.kt
sjoubert: Yes, except it does not test that. It tests the conflict between multipart file names (or dataset source names if I'm wrong. see my other comment). But it's not about the name of the dataset, right?

sjoubert (approved):

PR #1041, opened 2025-08-19
Jrey/split dataset list on runner

jreynard-code:
The aim of this PR is to split runner parameters in two separated properties:
- parameterValues will contain only "simple" parameters ( int, string, ...)
- complex type (file) will be stored as dataset part attached to a runner dataset (runner.datasets.parameter)
This will allow us to only have 1 dataset per runner (instead of 1 dataset per complex type)
- solution files endpoints have been removed (only used for default solution parameter with complex types) and now, we can override/set values for a solution parameter directly on workspace


runner/src/main/openapi/runner.yaml
```diff
@@ -1,8 +1,12 @@
...
 
 info:
-  title: ""
-  version: ""
+  title: Cosmo Tech Runner Manager API
+  description: Cosmo Tech Runner Manager API
+  version: 1.0.0-SNAPSHOT
+
+servers:
+  - url: 'http://localhost:8080'
```
sjoubert, line 9: Do we need that back? If I recall correctly, I removed all these from the sub-module openapi specs as they will be overwritten during the merge to generate the final spec. Before we had some warnings during the merge about overridden or ignored data that was fixed by not putting `info` and `servers` config in the split specs.

sjoubert (approved):

PR #1058, opened 2025-09-17
Make `files` parameter in dataset creation and update optional

jreynard-code:
- Updated `DatasetServiceImpl` to handle `files` parameter as optional in both dataset creation and update operations.
- Added integration tests to verify behavior when `files` parameter is omitted.
- Updated mustache template and OpenAPI generator configuration to reflect the optional nature of `files`.


jreynard-code:
OpenAPITools/openapi-generator PR : https://github.com/OpenAPITools/openapi-generator/pull/21994

PR #1059, opened 2025-09-25
merge common lib src to api src

lalepee:
I've pushed the src and do the linking on sub-modules


common/build.gradle.kts
```diff
@@ -0,0 +1,123 @@
...
+        constraints { implementation("com.redis:lettucemod:4.4.0") }
+      }
+  implementation("org.testcontainers:postgresql:${testContainersPostgreSQLVersion}")
+  implementation("org.testcontainers:localstack:${testContainersLocalStackVersion}")
+  implementation("org.apache.tika:tika-core:${tikaVersion}")
+  implementation("org.springframework.boot:spring-boot-starter-test")
+
+  testImplementation(kotlin("test"))
+  testImplementation(platform("org.junit:junit-bom:${jUnitBomVersion}"))
+  /*
```
sjoubert, line 54: Please remove this commented block if we actually don't need it.

sjoubert (changes requested):

lalepee:
SpotlessApply is also running fine

PR #1063, opened 2025-09-29
Remove vendor and identity provider code properties

jreynard-code:
- Eliminated `vendor` from `CsmPlatformProperties` and related configurations.
- Removed `identityProvider.code` from platform properties and associated usages in tests, configs, and runtime environments.
- Refactored environment profiles.


sjoubert (commented):
Does `PersistentMetric.kt` `vendor` have a link to the removed `vendor`?

PR #1157, opened 2026-03-23
add RunType to Run and start delete run on runner deletion

lalepee:


jreynard-code (approved):
Let's say it's kind of sufficient right now.
I think we have to take the subject from another point of view.
Let's try like this and see the limitation/usage after time

PR #1163, opened 2026-04-07
Spring boot 4 

jreynard-code:
## Spring Boot 4 Migration

Migrate the Cosmotech API from Spring Boot 3.x to **Spring Boot 4.0.6**, along with major dependency upgrades and test infrastructure modernization.

### Core Upgrades

- **Spring Boot** 3.x → 4.0.6
- **Gradle** → 9.4.1
- **Redis OM Spring** → 2.0.7 (with dependency resolution strategy enforcement)
- **OpenAPI Generator** → 7.23.0
- **SpringDoc** → 3.0.3
- **Swagger Parser** → 2.1.42
- **Jib Gradle Plugin** → 3.5.3
- **Base image** updated to `dhi.io/eclipse-temurin:21-alpine3.23`

### Build & Tooling

- Upgrade Axion Release, Spotless, Dependency Check, Kover, and CycloneDX BOM plugins
- Add Kotlin compiler argument to suppress warnings during compilation
- Update all CI/CD workflows (`.github/workflows/`) for Gradle 9 compatibility

### Test Infrastructure

- **Replace LocalStack with SeaweedFS** for S3-compatible storage in integration tests
- **Replace `RedisStackContainer` with `RedisContainer`** and upgrade testcontainers-redis to 2.2.4
- Replace deprecated `@SpykBean` with `@MockkSpyBean` across test suites
- Refactor `@BeforeEach` methods and remove unnecessary `flushall` calls for improved test reliability
- Initialize default S3 bucket for SeaweedFS in test base
- Remove unused imports and environment variables from test classes

### Application Changes

- Add **RedisOM map converters** (`RedisOMMapConverterConfig`) and pagination logic for dataset parts
- Update Redis serialization configuration
- Remove deprecated `SwaggerBeanConfiguration`
- Refactor `CsmApiConfiguration` and Keycloak security configuration for Spring Boot 4 compatibility
- Fix index annotation to OpenAPI schema on Solution
- Refactor RBAC tests and `DatasetServiceImpl` for cleaner code

### OpenAPI

- Add missing schema properties (accessControlList.role is Indexed now) in `dataset.yaml`, `workspace.yaml`, and `solution.yaml`


common/src/main/kotlin/com/cosmotech/common/redis/RedisOMMapConverterConfig.kt
```diff
@@ -0,0 +1,77 @@
...
+ * Redis OM Spring's MappingRedisOMConverter.writeMap() calls writeInternal() for map entry values,
+ * which then calls getRequiredPersistentEntity(LinkedHashMap.class) and fails.
+ *
+ * This fix patches the static converter list in RedisOMCustomConversions via reflection so that
+ * customConversions.hasCustomWriteTarget(LinkedHashMap.class) returns true, causing writeToBucket()
+ * to be called instead of writeInternal().
+ *
+ * BeanFactoryPostProcessor runs BEFORE any repository beans are instantiated, ensuring the
+ * converters are registered before SimpleRedisDocumentRepository creates its internal
+ * MappingRedisOMConverter.
```
sjoubert: Could you also add a link to the upstream issue you created for that?

common/src/main/kotlin/com/cosmotech/common/security/keycloak/KeycloakSecurityConfiguration.kt
```diff
@@ -171,7 +171,8 @@ class KeycloakJwtGrantedAuthoritiesConverter(
       if (rawRoleClaim is Collection<*>) {
         return rawRoleClaim
             .stream()
-            .map { role: Any? -> SimpleGrantedAuthority((role as? String)) }
+            .filter(Objects::nonNull)
+            .map { role -> SimpleGrantedAuthority((role as String)) }
             .filter(Objects::nonNull)
```
sjoubert: Shouldn't the second/original `nonNull` filter be useless now that we have one before the `map`. I'm assuming the map's lambda can't produce a null Object if it was filtered before.

sjoubert (changes requested):

sjoubert (approved):

api/src/integrationTest/kotlin/com/cosmotech/api/home/organization/OrganizationControllerTests.kt
```diff
@@ -377,18 +378,44 @@ class OrganizationControllerTests : ControllerTestBase() {
...
+        responseContent.get("status"),
+    )
+    assertEquals(
+        JSONObject(emptyNameOrganizationCreationRequestError).get("instance"),
+        responseContent.get("instance"),
+    )
+    assertEquals(
+        JSONObject(emptyNameOrganizationCreationRequestError).get("type"),
+        responseContent.get("type"),
+    )
```
lalepee: Why does it differs from other tests, should this method be copied on other tests ?

api/build.gradle.kts
```diff
@@ -21,7 +21,9 @@ dependencies {
   implementation(projects.cosmotechRunApi)
   implementation(projects.cosmotechRunnerApi)
   implementation(projects.cosmotechCommonApi)
-  testImplementation("org.springframework.security:spring-security-test")
+  testImplementation("org.springframework.boot:spring-boot-starter-security-test")
+  testImplementation("org.springframework.boot:spring-boot-restdocs")
```
lalepee, line 25: Tests needs  rest docs ???

common/src/main/kotlin/com/cosmotech/common/config/SwaggerBeanConfiguration.kt
lalepee, line 1: Is it replace by something ?

solution/src/main/openapi/solution.yaml
```diff
@@ -1208,6 +1208,7 @@ components:
           example: "user@company.com"
         role:
           type: string
+          x-field-extra-annotation: "@com.redis.om.spring.annotations.Indexed"
```
lalepee, line 1211: Index on the solution role ? Is it used ? if yes where ?

workspace/src/main/openapi/workspace.yaml
```diff
@@ -812,6 +812,7 @@ components:
           description: The identity id
         role:
           type: string
+          x-field-extra-annotation: "@com.redis.om.spring.annotations.Indexed"
```
lalepee, line 815: Same

dataset/src/main/kotlin/com/cosmotech/dataset/service/DatasetServiceImpl.kt
```diff
@@ -576,62 +577,16 @@ class DatasetServiceImpl(
...
-                    datasetId,
-                    pageable,
-                )
-                .toList()
-          }
+      val start = pageable.offset.toInt()
+      dataset.parts.subList(
+          min(start, dataset.parts.size),
+          min(start + pageable.pageSize, dataset.parts.size),
+      )
```
lalepee, line 588: This works but it asks the question why do we still have a datasetPartObject if it is stored as a whole in the datasetObject ?

lalepee (approved):
I have some inquiries. But LGTM

build.gradle.kts
```diff
@@ -641,13 +615,11 @@ tasks.getByName("spotlessKotlinGradle") {
 val copySubProjectsDetektReportsTasks =
     subprojects.flatMap { subProject ->
       listOf("html", "xml", "txt", "sarif").map { format ->
-        val formatCapitalized = format.capitalizeAsciiOnly()
+        val formatCapitalized = format
```
sjoubert: Seems you removed the capitalization of the format, so I guess we should just remove that intermediate variable and directly use `format` below.

sjoubert (changes requested):

api/build.gradle.kts
jreynard-code: It was already the case.
The API doc, with code snippets and more, is generated with all Controller tests that you can find under api/integrationTests.

common/src/main/kotlin/com/cosmotech/common/config/SwaggerBeanConfiguration.kt
jreynard-code: yep right here: common/src/main/kotlin/com/cosmotech/common/config/CsmApiConfiguration.kt
This bean was here to handle `octet-stream` has an acceptable mime-type.
With Spring boot 4, we have a bunch of converters available.
I set the octet-stream MIME-TYPE as supported mime-type on ResourceHttpMessageConverter (with is more relevant to me)

solution/src/main/openapi/solution.yaml
jreynard-code: In service layer.
Basically, when we're searching for resources (with RBAC applied) the FT.SEARCH query looks like this `.... -@security_accessControlList_role:{none}))` if we do not set the @Indexed annotation, this query is not valid anymore (I suspected a wrong behaviour in previous redisOM-spring lib that is now fixed)

workspace/src/main/openapi/workspace.yaml
jreynard-code: In service layer.
Basically, when we're searching for resources (with RBAC applied) the FT.SEARCH query looks like this `.... -@security_accessControlList_role:{none}))` if we do not set the @Indexed annotation, this query is not valid anymore (I suspected a wrong behaviour in previous redisOM-spring lib that is now fixed)

dataset/src/main/kotlin/com/cosmotech/dataset/service/DatasetServiceImpl.kt
jreynard-code: DatasetParts have their own lifecycle (CRUD method + index).
When we retrieve a Dataset from DB, we also retrieve the parts associated (define in the property parts of the Dataset). The point here is to not query again the DB to fetch data that we already have (normalized on Dataset object).
I could have call the method 
``` 
datasetPartRepository
                .findByOrganizationIdAndWorkspaceIdAndDatasetIdNoSecurity(
                    organizationId,
                    workspaceId,
                    datasetId,
                    pageable,
                )
                .toList()
```

instead of dealing directly with the parts property but it's straightforward to do it :sweat_smile:

sjoubert (approved):

PR #1172, opened 2026-05-27
fix execution ordre lauch, wait then delete

lalepee:
I was forced to add a waiting loop on the deletion run. I did not find a better way.It's ugly but it works


jreynard-code:
I made some changes:
- replaced match { } in tests by any() calls
- added the configuration associated to your cleanUpArchivedRunners task: 
  - now we can disable it and change the delay in configuration
  - the task is **enabled** by default 
  - default delay is **300s**

**Global remarks:**
- following points need to be documented and listed in the next release notes (could you please check with doc team and our PO):
  - task configuration
  - status on Runner and meanings
  - runner archive flow and runner cleanUp flow
- As discussed, it's a hot fix. we need to define together a proper way to handle the output deletion.

jreynard-code (approved):
Looks good enough to me 👍 

See my comment for further details ❤️

PR #1201, opened 2026-07-20
Update openapi-generator to 7.24

sjoubert:
In particular, use the 'quiet' option to disable the banner display

Proper tests, in particular for the clients, still need to be done.
You might also want to check if the JsonNullable/optional property is correctly implemented/fixed and remove the template workaround we have for that.


jreynard-code (approved):
LGTM

PR #1203, opened 2026-07-27
Lal/keycloak group members prod 16016

lalepee:
This add a link from the API to Keycloak admin which is a different link than the authentication.


organization/src/main/kotlin/com/cosmotech/organization/service/OrganizationServiceImpl.kt
```diff
@@ -74,6 +82,21 @@ class OrganizationServiceImpl(
     return result
   }
 
+  override fun listKeycloakGroups(): List<String> {
+    // open to all users, no permission check needed
+    return keycloak.getAllGroups().map { it.name }
+  }
+
+  override fun listKeycloakMembers(): OrganizationMembers {
```
sjoubert: According to the design, shouldn't this be secured behind a `Platform.Admin`?

organization/src/main/openapi/organization.yaml
```diff
@@ -83,6 +83,62 @@ paths:
                 TwoOrganizations:
                   $ref: '#/components/examples/TwoOrganizations'
 
+  /organizations/members:
+    get:
+      operationId: listKeycloakMembers
```
sjoubert: * Should we avoid mentioning keycloak explicitly in the endpoint name? I'm not sure how much the logic and response data is specific to that IAM, but it could make more sense to name them `listMembers` (although a bit too generic), `listIAMMembers`, or something else.
* Are we ok with using the Organization group to "store" these 2 endpoints even though they are not specific to organizations. We have a Meta group that could store those (directly or under a sub-path) or we could create its own module (maybe that's more work than needed)

common/src/main/kotlin/com/cosmotech/common/security/keycloak/KeycloakClient.kt
```diff
@@ -0,0 +1,140 @@
...
+    return userResource.roles().realmLevel().listAll()
+  }
+
+  fun getAllUsers(): List<UserRepresentation> {
+    val keycloak = getKeycloakInstance()
+    val realmResource = keycloak.realm(realm)
+    return realmResource.users().list()
+  }
+
+  fun listCosmotechMembers(rbac: List<RbacAccessControl>): KeycloakMembers {
```
sjoubert: I'm not sure of the meaning of `Cosmotech` here, what are Cosmotech's members compared to regular members? I understand this is a rbac-filtered list, maybe we can find a better name for the method.

common/src/main/kotlin/com/cosmotech/common/security/keycloak/KeycloakClient.kt
```diff
@@ -0,0 +1,140 @@
...
+            },
+    )
+  }
+
+  private fun extractKeycloakRole(roles: List<String>?): String {
+    roles ?: return "None"
+    return when {
+      roles.contains("Platform.Admin") -> "Platform.Admin"
+      roles.contains("Organization.User") -> "Organization.User"
+      else -> "None"
```
sjoubert: * Should we use the global variables defined in AbstractSecurityConfiguration instead of hard-coded literals?
* And what about `Organization.Viewer`? Maybe it's not really used and not worth dealing with?

organization/src/main/openapi/organization.yaml
```diff
@@ -591,6 +647,55 @@ components:
       required:
         - role
 
+    OrganizationMembers:
```
sjoubert, line 614: Mentioning it once, but I think it applies to all modules:
- `users` and `groups` should be required, and have a default `[]` value to avoid the field being treated as nullable
- Same for the `users` in MemberGroups
- As for the documentation mentionning "default to null" on `id` and `role` I don't think we can do much, this is a current limitation of the generator.

sjoubert (changes requested):
I think we also need to add tests for all the new endpoints:
* unit/integration tests that check the RBAC requirements
* controller test in the main root module to generate documentation

The merge will also be blocked with the integration of the new keycloak configuration into terraform (maybe the helm chart, but I don't think so) and by the deployment/update of tenant (and the broken CI tenant that needs to be moved from warp to vela) with the new terraform module.

common/src/main/kotlin/com/cosmotech/common/tests/CsmTestBase.kt
```diff
@@ -104,5 +128,6 @@ open class CsmTestBase {
     postgres.stop()
     seaweedServer.stop()
     redisServer.stop()
+    keycloakServer.stop()
```
sjoubert, line 131: Should the keycloak container be stopped first? All the other containers are stopped in reverse starting order. I don't think there are much dependencies between them so it should not matter much.

organization/src/main/kotlin/com/cosmotech/organization/service/OrganizationServiceImpl.kt
sjoubert: 🤔 Actually I'm not sure the security check should be done here. Shouldn't it be handled somewhere in `common/src/main/kotlin/com/cosmotech/common/security/AbstractSecurityConfiguration.kt`? As well as all the other new endpoints?

sjoubert (changes requested):

build.gradle.kts
```diff
@@ -335,6 +336,8 @@ subprojects {
 
     implementation(platform("io.awspring.cloud:spring-cloud-aws-dependencies:$awsSpringVersion"))
     implementation("io.awspring.cloud:spring-cloud-aws-starter-s3:$awsSpringVersion")
+    // Source: https://mvnrepository.com/artifact/org.keycloak/keycloak-admin-client
+    implementation("org.keycloak:keycloak-admin-client:26.0.10")
```
jreynard-code: Could you please extract version number into a variable and remove the comment above?

workspace/src/test/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImplTests.kt
```diff
@@ -109,6 +110,8 @@ class WorkspaceServiceImplTests {
 
   @InjectMockKs private lateinit var workspaceServiceImpl: WorkspaceServiceImpl
 
+  @Suppress("unused") @InjectMockKs private lateinit var keycloakClient: KeycloakClient
```
jreynard-code: Do we need that?

common/src/main/kotlin/com/cosmotech/common/security/AbstractSecurityConfiguration.kt
```diff
@@ -155,7 +155,7 @@ internal fun endpointSecurityReaders(
 ) =
     listOf(
         CsmSecurityEndpointsRolesReader(
```
jreynard-code: Without any other definition, every user with ORGANIZATION_USER or ORGANIZATION_VIEWER will have access to /members. Do we really want that?
If yes, please add specific /members in each PATHS_* variable to explicitly define it.
If not, we need to define the right configuration on each /members endpoints

api/src/integrationTest/kotlin/com/cosmotech/api/home/iaminfo/IAMInfoControllerTests.kt
```diff
@@ -0,0 +1,69 @@
...
+
+  private val logger = LoggerFactory.getLogger(IAMInfoControllerTests::class.java)
+
+  @Test
+  fun `list IAM groups as platform admin`() {
+    mvc.perform(get("/iaminfo/groups").withPlatformAdminHeader().accept(MediaType.APPLICATION_JSON))
+        .andExpect(status().is2xxSuccessful)
+        .andExpect { result ->
+          // response should be a JSON array of group names, open to all users
+          JSONArray(result.response.contentAsString)
```
jreynard-code: Could you please add response data checks ?
Same for all tests in this class

iaminfo/src/main/kotlin/com/cosmotech/iaminfo/service/IAMInfoServiceImpl.kt
```diff
@@ -0,0 +1,51 @@
...
+    private val keycloak: KeycloakClient,
+    private val csmAdmin: CsmAdmin,
+) : CsmPhoenixService(), IAMInfoApiServiceInterface {
+  override fun listIAMGroups(): List<String> {
+    // open to all users, no permission check needed
+    return keycloak.getAllGroups().map { it.name }
+  }
+
+  override fun listIAMMembers(): Members {
+    // only be accessible by Platform.admin users
```
jreynard-code: It should also be secured in AbstractSecurityConfiguration

iaminfo/src/integrationTest/resources/init-db.sql
```diff
@@ -0,0 +1,2 @@
+CREATE USER readusertest WITH PASSWORD 'readusertest';
+CREATE USER writeusertest WITH PASSWORD 'writeusertest';
```
jreynard-code: is it needed for something?

common/src/main/kotlin/com/cosmotech/common/security/keycloak/KeycloakClient.kt
```diff
@@ -0,0 +1,144 @@
...
+    val serverUrl: String = csmPlatformProperties.identityProvider.serverBaseUrl,
+    val realm: String = csmPlatformProperties.identityProvider.identity.tenantId,
+    val adminClientId: String = csmPlatformProperties.identityProvider.admin.clientId,
+    val username: String = csmPlatformProperties.identityProvider.admin.username,
+    val password: String = csmPlatformProperties.identityProvider.admin.password,
+) {
+  companion object {
+    const val MAX_PAGE_SIZE = 100
+    const val OFFSET_START = 0
+  }
```
jreynard-code: Could you please move it to a Constant class?
Below, you can also set all hardcoded string values (count, public, *) in the Constant class

common/src/main/kotlin/com/cosmotech/common/security/keycloak/KeycloakClient.kt
```diff
@@ -0,0 +1,144 @@
...
+                  id = group.name,
+                  role = extractKeycloakRole(group.realmRoles),
+                  users = getUsersInGroup(group.id).map { it.username }.distinct(),
+              )
+            },
+    )
+  }
+
+  private fun extractKeycloakRole(roles: List<String>?): String {
+    roles ?: return "None"
```
jreynard-code: "None" can be replaced by com.cosmotech.common.rbac.ROLE_NONE

common/src/test/kotlin/com/cosmotech/common/security/keycloak/KeycloakClientTests.kt
```diff
@@ -0,0 +1,463 @@
...
+    keycloakInstance = mockk(relaxed = true)
+    realmResource = mockk(relaxed = true)
+    groupsResource = mockk(relaxed = true)
+    usersResource = mockk(relaxed = true)
+
+    every { keycloakClient.getKeycloakInstance() } returns keycloakInstance
+    every { keycloakInstance.realm(any()) } returns realmResource
+    every { realmResource.groups() } returns groupsResource
+    every { realmResource.users() } returns usersResource
+  }
```
jreynard-code: It can be rewrote as following:

```
    val keycloakClient = spyk(KeycloakClient(
      csmPlatformProperties = mockk(),
      serverUrl = "http://localhost:8080",
      realm = "cosmotech",
      adminClientId = "admin-cli",
      username = "admin",
      password = "admin",
    ))

  // Keycloak admin client chain mocks
  val keycloakInstance: Keycloak = mockk()
  val realmResource: RealmResource = mockk()
  val groupsResource: GroupsResource = mockk()
  val usersResource: UsersResource = mockk()

  @BeforeEach
  fun setUp() {
    every { keycloakClient.getKeycloakInstance() } returns keycloakInstance
    every { keycloakInstance.realm(any()) } returns realmResource
    every { realmResource.groups() } returns groupsResource
    every { realmResource.users() } returns usersResource
  }```

File: common/src/test/kotlin/com/cosmotech/common/security/keycloak/KeycloakClientTests.kt [hunk 16]
```diff
@@ -0,0 +1,463 @@
...
+import org.keycloak.admin.client.resource.RealmResource
+import org.keycloak.admin.client.resource.RoleMappingResource
+import org.keycloak.admin.client.resource.RoleScopeResource
+import org.keycloak.admin.client.resource.UserResource
+import org.keycloak.admin.client.resource.UsersResource
+import org.keycloak.representations.idm.GroupRepresentation
+import org.keycloak.representations.idm.RoleRepresentation
+import org.keycloak.representations.idm.UserRepresentation
+
+class KeycloakClientTests {
```
jreynard-code: You can remove all `relaxed = true`, there are useless in this test class

jreynard-code (changes requested):
Some request changes from my desk with luv'

workspace/src/test/kotlin/com/cosmotech/workspace/service/WorkspaceServiceImplTests.kt
lalepee: oui

common/src/main/kotlin/com/cosmotech/common/security/AbstractSecurityConfiguration.kt
lalepee: I've added a /members on each modules

iaminfo/src/integrationTest/resources/init-db.sql
lalepee: Yes, for the initialization of the CsmTestBase  Redis database. Like in every other integration test for the other modules.

iaminfo/src/main/kotlin/com/cosmotech/iaminfo/service/IAMInfoServiceImpl.kt
lalepee: done

common/src/main/kotlin/com/cosmotech/common/security/keycloak/KeycloakClient.kt
lalepee: I'm not sure, this is not an RBAC role but a Kaycloak Realm Role. So .....

api/src/integrationTest/kotlin/com/cosmotech/api/home/iaminfo/IAMInfoControllerTests.kt
lalepee: done

common/src/test/kotlin/com/cosmotech/common/security/keycloak/KeycloakClientTests.kt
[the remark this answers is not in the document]
lalepee: I've kept only the section delimitation Tests Helpers

common/src/main/kotlin/com/cosmotech/common/security/keycloak/KeycloakClient.kt
lalepee: Return empty string in case of not role define in Keycloak.

jreynard-code (approved):
LGTM ;)

PR #1213, opened 2026-08-10
Fix typescript generator due to an incompatibility with axios 1.19

sjoubert:
This will be fixed in openapi-generator 7.25
See https://github.com/OpenAPITools/openapi-generator/issues/24523


jreynard-code (approved):
LGTM

